PR #29, opened 2019-05-01
Bookdown setup

lwjohnst86:
Here's the basic bookdown with travis setup. Still working out why I can't get travis to see my admin privileges for this repo, but this should work when it is fixed.


lwjohnst86:
Nice, got it to start building via travis. We'll see how it deploys.

PR #45, opened 2019-05-07
Starting to build internal consistency checks.

gvwilson:
-   Add scripts `bin/checkgloss.py` and `bin/checklinks.py` to check glossary entries and external links.
-   Add `bin/util.py` for common support.
-   Add target to `Makefile`.
-   Update `.gitignore` to ignore `__pycache__` generated by loading `util.py`.
-   Removed unused links from `etc/links.md`.

These scripts produce false warnings because they cannot distinguish `a[0][1]` from link usage, and because the example of how to write a glossary entry looks like a glossary entry.


ljdursi:
Cool, I'll take a look.

lwjohnst86:
@gvwilson if you wanna fix the conflict, you can merge it in.

lwjohnst86:
Question, will we need to run these ourselves or they are automatic? If not, how do we use these?

gvwilson:
Manually for the foreseeable future, then w/ Travis as we get closer to publication. Please don't merge - I need to do some more work to bring these up to date.

PR #62, opened 2019-05-15
Adding novels from Project Gutenberg as a possible dataset for textual analysis projects.

gvwilson:
-   former `project` directory renamed to `zipfs-law`.
-   `.gitignore` updated to ignore the `zipfs-law/results` directory.
-   Added `zipfs-law/data/README.md` (simple data manifest).


elliewix:
I have used their copy of Dracula with my class for about three years now and know a bunch of interesting things you can analyze with that text. I also have a bunch of exercises related to it.

Also I don't know how to do reviews in github.

gvwilson:
I think a comment like the one you just left is the kind of review something like this needs :-)

ChristinaLK:
This looks cool! I have no idea what exercises I would use from it, but it sounds like @elliewix has that covered. ;)

gvwilson:
@lwjohnst86 please merge - I think this is good to go.

PR #63, opened 2019-05-15
Outline for novice material + some formative assessments for novice R

gvwilson:
This doesn't cover everything learners should be able to do by the time they're finished the novice R material, but I think they should be able to do at least this much - please add comments on what you think is out of scope and what else you think ought to be there.


gvwilson:
I have added another file `novice-goals.Rmd` with a point-form list of things I think both novice courses might cover - like the R exercises, it's meant solely as a starting point for discussion, and I'd be very grateful for comments, fixes, additions, or, "Yo, Greg, wuzzis?"

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+4. Colorize scatterplot points by class:
+
+```{r colorize-scatterplot}
+hra %>%
+  dplyr::mutate(class_fct = as.factor(class)) %>%
+  ggplot2::ggplot() +
+  ggplot2::geom_point(mapping = aes(x = log10.mass, y = log10.hra, color = class_fct), alpha = 0.5)
+```
+
+**Group by order and experiment with different alpha values.*
```
k8hertweck: missing * and everything is SO BOLD

novice-goals.Rmd
```diff
@@ -0,0 +1,201 @@
...
+-   What tests should I write for my software?
+    -   edge cases
+    -   fails when it should
+    -   bugs become tests
+-   What *isn't* included?
+    -   legal liabilities
+
+## Version control
+
+-   What is a version control system?
```
k8hertweck: explicitly including remote vs local, as this is both a conceptual (e.g., "where is my stuff?") and ethical (e.g., "are these PHI data safe?") issue

novice-goals.Rmd
```diff
@@ -0,0 +1,201 @@
...
+    -   variable expansion
+-   What *isn't* included?
+    -   templating
+    -   filters
+    -   inclusions
+
+## Reproducibility
+
+-   How can I make programs easy to read?
+    -   coding style
```
k8hertweck: linters?

novice-goals.Rmd
```diff
@@ -0,0 +1,201 @@
...
+    -   build tools (Make and its kin)
+    -   continuous integration
+    -   documentation generators
+
+## Collaboration
+
+-   What kinds of licenses are there?
+    -   open vs. closed
+    -   copyright
+-   What license should I use for my publications?
```
k8hertweck: Even before this, "Who gets to decide what license to use?" since some organizations, granting agencies, etc may have restrictions

novice-goals.Rmd
```diff
@@ -0,0 +1,201 @@
...
+    -   DOIs
+    -   ORCIDs
+-   What's the difference between open and welcoming?
+    -   evidence for systematic exclusion
+    -   mechanics of exclusion
+-   How can I help create a level playing field?
+    -   what's wrong with deficit models
+    -   allyship
+    -   Code of Conduct (remove negatives)
+    -   kerb cuts (add positives)
```
k8hertweck: Yo, Greg, wuzzis?

k8hertweck:
Added a few comments to the outline for minor points, still thinking about bigger picture.

novice-r-practice.Rmd
gvwilson: **FIXED**

novice-goals.Rmd
gvwilson: thanks

novice-goals.Rmd
gvwilson: good catch

novice-goals.Rmd
gvwilson: :+1:

novice-goals.Rmd
gvwilson: switched to simplified (American) spelling and added a link

novice-goals.Rmd
```diff
@@ -0,0 +1,207 @@
...
+    -   CC0
+-   How should I identify myself and my work?
+    -   DOIs
+    -   ORCIDs
+-   What's the difference between open and welcoming?
+    -   evidence for systematic exclusion
+    -   mechanics of exclusion
+-   How can I help create a level playing field?
+    -   what's wrong with deficit models
+    -   allyship
```
brandeism: add advocacy and sponsorship

novice-goals.Rmd
```diff
@@ -0,0 +1,207 @@
...
+    -   assertions
+-   How can I write and manage tests for my software?
+    -   unit tests
+    -   test runners
+-   What tests should I write for my software?
+    -   edge cases
+    -   fails when it should
+    -   bugs become tests
+-   What *isn't* included?
+    -   legal liabilities
```
brandeism: add - Has this program provided the answers we are seeking? What are our conclusions?

novice-goals.Rmd
```diff
@@ -0,0 +1,207 @@
+# Outline {#outline}
+
+This outline describes the questions that the novice courses on R and Python will answer.
+The advanced course can then assume that learners have hands-on experience with these topics
+but nothing more.
+
```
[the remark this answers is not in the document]
gvwilson: The first two look great to me - what questions do you have students investigate with the Louisville data?

novice-goals.Rmd
```diff
@@ -0,0 +1,207 @@
...
+    -   In the IDE
+    -   Stack Overflow
+-   How can I ask for help?
+    -   creating a reproducible example (reprex)
+-   How do I install software?
+    -   package manager
+-   How can I inspect data while I'm working on it?
+    -   table viewers
+-   How can I tell what's gone wrong in my programs?
+    -   reading error messages
```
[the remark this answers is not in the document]
gvwilson: That's a good point - I'll move it up.

novice-goals.Rmd
```diff
@@ -0,0 +1,207 @@
...
+    -   3D visualization
+
+## Data analysis
+
+-   How can I represent and manage missing values?
+    -   NA
+-   How can I get a feel for my data?
+    -   summary statistics
+-   How can I create a simple model of my data?
+    -   formulas
```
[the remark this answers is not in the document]
gvwilson: Is k-means cluster analysis enough for this section? (I've added a line...)

novice-goals.Rmd
gvwilson: Done - thanks.

novice-goals.Rmd
gvwilson: Do you mean "add to the list of things we're not including in the novice course" or "add as a major point that we should cover"? I'm assuming the former...

novice-goals.Rmd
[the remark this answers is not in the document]
brandeism: 1. How do salaries change from year to year?
2. Is the a relationship amongst the police officers' salaries (RegularRate, OvertimeRate, IncentiveAllowance, Other)

We didn't get to track the salary of a single employee across years, but I think it would be cool to incorporate it, if possible.

novice-goals.Rmd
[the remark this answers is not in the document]
brandeism: That should work.

novice-goals.Rmd
brandeism: Yes, I meant to add it to the former.

novice-goals.Rmd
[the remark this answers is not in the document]
gvwilson: :+1:

novice-goals.Rmd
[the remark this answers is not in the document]
gvwilson: :+1:

novice-goals.Rmd
gvwilson: :+1:

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   Abela's Rules
+-   How can I create different kinds of plots?
+    -   scatter plot
+    -   line plot
+    -   histogram
+-   How can I plot multiple datasets at once?
+    -   grouping
+    -   faceting
+-   How can I make misleading plots?
+    -   examples of what not to do
```
joelostblom: - Showing a single central tendency data point instead of the individual observations
    - And the opposite, saturated plots instead of for example violins or 2D histograms
- Not start value axis at zero to exaggerate a difference
- Not using perceptually uniform colormaps to indicate quantities (e.g. rainbow ones)
    - And not thinking about color blindness (e.g. also map shape or use appropriate colors).

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+-   What *isn't* included?
```
joelostblom: outliers

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+## Plotting
+
+-   What are the core elements of every plot?
+    -   data
+    -   geometry
+    -   mapping
+-   What kinds of plot should I use?
+    -   Abela's Rules
+-   How can I create different kinds of plots?
```
joelostblom: Can we include some brief theory here on plot groupings (e.g. categorical vs quantitative, quantitative vs quantitative, and distributions/counts).

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   anything other than reasonably tidy tabular data
+    -   map
+    -   loops and conditionals
+
+## Programming
+
+-   What are the different ways I can interact with software?
+    -   console
+    -   scripts
+    -   libraries/packages?
```
joelostblom: maybe mention libraries as a reusable extension of scripts/functions

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   linear regression
+    -   k-means cluster analysis
+-   How can I put people at risk?
+    -   algorithmic bias
+    -   de-anonymization
+-   What *isn't* included?
+    -   statistical tests
+    -   anything with "machine learning" in its name
+
+## Verification
```
joelostblom: IMO, this is a more advanced section, could it be moved to intermediate? Possibly replaced with elaboration on some of the concepts in the previous headings to solidify learners understanding of programmatic data analysis before moving on to the parts outside the analysis itself.

joelostblom: An possible replacement could be an example EDA from reading in the data to making some basic plots. Just the first steps that should nearly always be done (e.g. read_csv, head, struct/glimpse, overview plots of variable distributions and pairwise relationships). So putting together what they have learnt, but with less details and more realistic flow.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+## Data analysis
+
+-   How can I represent and manage missing values?
+    -   NA
+-   How can I get a feel for my data?
+    -   summary statistics
+-   How can I create a simple model of my data?
+    -   formulas
+    -   linear regression
```
cwickham: At what level are you thinking for linear regression here?  
* Adding a "best fit" straight line on a scatterplot - this is within scope for sure.
* Understanding what the error bands on a "best fit" straight line mean - perhaps...I'd like novices to understand what kind of uncertainty these attempt to represent...but they aren't appropriate in so many situations. 
* Multiple linear regression - I'm not so sure.  Yes, I think this is important, but requires a big cognitive leap to understand the connection between the model and the data (e.g. the whole "holding everything else constant" part makes it much harder to visualize the model and data in one plot).

novice-goals.Rmd
cwickham: I'm in agreement with Joel here.  It feels like everything up to this point has been very much "code as you go" data analysis, and now these topics feel like we are jumping to automating the process.  

I think this would be a good place to move some of the Reproducibility topics.  E.g. "How can I combine explanations, code, and results?" and move focus to putting together a start to finish analysis.  Testing in this context might be less formal than unit tests and involve data summaries, tables or plots that the reader/analyst can use to verify things went as planned.

novice-goals.Rmd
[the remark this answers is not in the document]
cwickham: We might frame these as exploratory tools for revealing structure in the data, rather than modelling or inferential tools.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   build tools (Make and its kin)
+    -   continuous integration
+    -   documentation generators
+
+## Collaboration
+
+-   What kinds of licenses are there?
+    -   open vs. closed
+    -   copyright
+-   Who gets to decide what license to use?
```
cwickham: Add: How do I credit someone else's code?  E.g Citing packages, citing something from someone else's GitHub, giving credit for someone's answer on StackOverflow.

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+*Create a table with only the latitudes and longitudes of sites south of -48 degrees.*
+
+9. Use `dplyr::mutate` to create a new column with calculated values and `stringr::str_length` to calculate string length:
+
+```{r mutate-name-length}
+person %>%
+  dplyr::mutate(name_length = stringr::str_length(family_name))
+```
+
+*Use the built-in function `round` to create a table with latitudes and longitudes rounded to integers.*
```
cwickham: Would students already have seen `round()`?  Or is this also an exercise in figuring out how to use a "new to you" function, e.g. with `?round`?

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+## Data manipulation
+
+-   How can I read tabular data into a program?
+    -   CSV
+    -   reading files
+-   How can I select subsets of my data?
+    -   select
+    -   filter
+    -   Boolean conditions
```
cwickham: Does arrange fit in this section somewhere?

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+1. Read `survey/measurements.csv` and look at the data with `View`:
+
+```{r read-and-view}
+measurements <- readr::read_csv(here::here("novice-r-practice", "measurements.csv"))
+View(measurements)
+```
+
+2. Find rows where `reading` is not NA and saved as `cleaned`:
+
```
cwickham: Report how many rows were removed. E.g. `nrow(measurements) - nrow(cleaned)`

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+*Create a plot faceted by order for just the reptiles.*
+
+6. Fit a linear regression to the logarithmic data for birds:
+
+```{r fit-line}
+hra %>%
+  dplyr::filter(class == "aves") %>%
+  ggplot2::ggplot() +
+  ggplot2::geom_point(mapping = aes(x = log10.mass, y = log10.hra), alpha = 0.5) +
+  ggplot2::geom_smooth(method = lm, mapping = aes(x = log10.mass, y = log10.hra), color = 'red')
```
cwickham: I advocate for `se = FALSE` in almost all situations

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+9. Save the linear regression plot for birds as a PNG:
+
+```{r save-file}
+hra %>%
+  dplyr::filter(class == "aves") %>%
+  ggplot2::ggplot() +
+  ggplot2::geom_point(mapping = aes(x = log10.mass, y = log10.hra), alpha = 0.5) +
+  ggplot2::geom_smooth(method = lm, mapping = aes(x = log10.mass, y = log10.hra), color = 'red')
+ggsave("/tmp/birds.png")
```
cwickham: Use `here()` instead of fixed path?

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+  dplyr::group_by(site_id) %>%
+  dplyr::mutate(delta_rad = reading - dplyr::lag(reading)) %>%
+  dplyr::filter(!is.na(delta_rad)) %>%
+  dplyr::summarize(any_increase = any(delta_rad > 0)) %>%
+  dplyr::filter(any_increase)
+```
+
+*Find sites with visits more than one year apart.*
+
+## Creating charts
```
cwickham: Nothing in this plotting section relies on previous data manipulation sections.  I'd love to see visualization first, then exercises above could use plots to verify results.

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
+# Practice
```
cwickham: These all look great!  Agree they are all "must include"s.

novice-r-practice.Rmd
```diff
@@ -0,0 +1,363 @@
...
+```
+
+*Do the same calculation use `case_when`.*
+
+6. Read `visited.csv`, drop the NAs, and join with the cleaned-up table of readings:
+
+```{r join-two-tables}
+cleaned <- readr::read_csv(here::here("novice-r-practice", "visited.csv")) %>%
+  dplyr::filter(!is.na(visit_date)) %>%
+  dplyr::inner_join(cleaned, by = c("visit_id" = "visit_id"))
```
cwickham: How much coverage of joins do you think is appropriate?  I.e. would a learner use `inner_join()` here because it is the only one they know about? Or is choosing an appropriate join part of this exercise?

A very useful skill is being able to see which rows don't join, i.e `anti_join()`, particularly when joining on columns that are subject to human data entry error.

cwickham:
Also just occurred to me I didn't see anywhere in `novice-goals.Rmd` any high-level "Where does stuff live on my computer?" topics.  Perhaps this is part of "File import" or "Project Organization", but we should be explicit about covering things like the structure of the file system, relative and absolute paths, and the difference between data living in files on the hard disk versus in objects in memory.

joelostblom:
I addition to what Charlotte mentioned in her comment above, I think it would be helpful with a general intro to computing. I am thinking brief and basic just to confer the mindset of doing things programatically instead of via point and click. Similar to [this "Working with computers" section](https://uoftcoders.github.io/rcourse/lec01-introduction.html#working_with_computers).

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
+# Outline {#outline}
+
+This outline describes the questions that the novice courses on R and Python will answer.
+The advanced course can then assume that learners have hands-on experience with these topics
+but nothing more.
+
+## Data manipulation
+
+-   How can I read tabular data into a program?
+    -   CSV
```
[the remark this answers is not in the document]
gvwilson: It does now - thanks.

novice-goals.Rmd
gvwilson: good catch - thanks

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   select
+    -   filter
+    -   Boolean conditions
+-   How can I calculate new values?
+    -   mutate
+    -   ifelse
+-   How can I tell what's gone wrong in my programs?
+    -   reading error messages
+    -   the difference between syntax and runtime errors
+-   How can I operate on subsets of my data?
```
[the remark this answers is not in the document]
gvwilson: Added - thanks.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   group
+    -   summarize
+-   How can I work with two or more datasets?
+    -   join
+-   What *isn't* included?
```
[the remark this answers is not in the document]
gvwilson: good catch - added.

novice-goals.Rmd
gvwilson: I don't think we can do that yet - they haven't seen how to write functions of their own.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   libraries/packages?
+-   How can I find and view help?
+    -   In the IDE
+    -   Stack Overflow
+-   How can I ask for help?
+    -   creating a reproducible example (reprex)
+-   How do I install software?
+    -   package manager
+-   How can I inspect data while I'm working on it?
+    -   table viewers
```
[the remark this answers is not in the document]
gvwilson: Good catch - moved up to the start of data manipulation.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+-   What *isn't* included?
+    -   code browsers, multiple cursors, and other fancy IDE tricks
+    -   virtual environments
+    -   debuggers
+
+## Plotting
+
```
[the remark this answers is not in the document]
gvwilson: Good - added. I also moved plotting up before programming.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+-   What are the core elements of every plot?
+    -   data
```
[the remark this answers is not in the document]
gvwilson: I think that would confuse more than illuminate at this level but I will defer to @cwickham.

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   geometry
```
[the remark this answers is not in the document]
gvwilson: fixed

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+    -   mapping
```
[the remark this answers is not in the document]
gvwilson: Will novices run into any kind of mapping that *isn't* aesthetic?

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+-   What kinds of plot should I use?
+    -   Abela's Rules
```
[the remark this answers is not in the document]
gvwilson: Added link to http://extremepresentation.typepad.com/files/choosing-a-good-chart-09.pdf - I find this decision tree a great way to explain how to choose a chart, and have permission to re-use the diagram (with citation).

novice-goals.Rmd
gvwilson: @cwickham what do you think?

novice-goals.Rmd
```diff
@@ -0,0 +1,210 @@
...
+-   How can I create different kinds of plots?
+    -   scatter plot
+    -   line plot
+    -   histogram
```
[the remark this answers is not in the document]
gvwilson: added - thank you.

novice-goals.Rmd
gvwilson: added

novice-goals.Rmd
gvwilson: added

novice-goals.Rmd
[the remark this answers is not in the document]
gvwilson: added - please make sure I understood correctly?

novice-goals.Rmd
gvwilson: 1.  If we're going to move unit testing to the intermediate material, I would still like to explain assertions and `stopifnot` and similar sanity checking tools here as part of reproducibility.
2.  If we're going to do that, then I think the section labelled "Programming" should start by showing people how to write functions (which ought to be more prominent anyway, now that I look at it), and including a few assertions or checks can then be discussed as part of good style for writing functions.

Thoughts?

novice-goals.Rmd
gvwilson: good catch - added.

novice-r-practice.Rmd
gvwilson: Good catch - exercise modified to ask them to look at the help first.

novice-r-practice.Rmd
gvwilson: added - thanks.

novice-r-practice.Rmd
gvwilson: Cool - can you please suggest an exercise appropriate to this level?

novice-r-practice.Rmd
gvwilson: I think we'd need a larger dataset for that - this one is fabricated and only has a handful of entries. (I wanted to keep it small so that learners could hand-check results, but maybe that's a mistake?)

novice-r-practice.Rmd
gvwilson: Happy to add - what's the reason (and why is `se = TRUE` the default)?

novice-r-practice.Rmd
gvwilson: good catch - thanks.

gvwilson:
I've also added some stuff on where files live, paths, and memory vs. disk - I think "working with computers" is a great idea, but I don't know where it should go - lead off with that? @joelostblom can you please send bullet points and some idea of where they should live and I'll update the PR?

novice-goals.Rmd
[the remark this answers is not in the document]
cwickham: Someone might argue that novices might confuse "mapping" with putting data on a map...although I'm not sure I've come across this confusion.

I mostly avoid "aesthetic" only because not that many people have encountered the word before.  I tend to phrase it something like "mapping between data variables and the visual properties of geometric objects".  

But, it inevitably comes up when explaining `aes()`, and "aesthetic mappings" is a reminder to always wrap in `aes()` (unless you are setting visual properties).

novice-goals.Rmd
cwickham: I'm a fan of getting novices familiar with a few common plot forms (e.g. those built on points, lines and bars) and a small number of general guidelines.  My current thinking is the most useful advice is "important comparisons are easiest when they are on the same scale and close together".

I think "Abela's Rules" looks a bit too complex for a novice (and I disagree strongly with some suggestions, e.g. a 3D area chart should never be used).  I do however like the idea of focusing on purpose - "What would you like to show?".

I also like the approach in [Fundamentals of Data Visualization](https://serialmentor.com/dataviz/directory-of-visualizations.html) where the focus is on what you are visualizing (amounts, proportions, trends, distributions, associations etc.), but again this probably goes into way more detail and options than a novice needs.

The breakdown in R4DS's [Exploratory Data Analysis](https://r4ds.had.co.nz/exploratory-data-analysis.html) of questions into those about variation and co-variation also works quite well.

So, I guess after introducing some common forms, I'd like to see us give some guidance of choosing plots to match questions of interest - with the emphasis being first on "what are you trying to show/see?", and then broken down into a few strategies that depend on the types of variables involved.

Will we need to teach these learners what to look for as well?  I.e. can we assume they know how to read a scatterplot?

novice-r-practice.Rmd
cwickham: I think inside a single lesson small hand-checkable examples are great!  And these are perfect for that.  

I was just thinking for exercises that synthesize over a few lessons, it would be nice to have tasks that pull from earlier sections, and to emphasize visualization as a tool throughout the analysis process, not just at the end.

novice-r-practice.Rmd
cwickham: Ah now you've forced me to articulate it...I've been flip-flopping for an hour!  Here are my thoughts:

Argument against including bands by default:

The bands answer an inferential question, that is, a question beyond the data at hand.  For us to even ask such a question we need some mechanism for connecting the "data we have", to the "data we *could have* had".  Then, for these bands to be appropriate that mechanism needs to match that assumed in their calculation.  Or, in other words, there are a **ton** of assumptions that go into their calculation and we should not treat those lightly.   We might also ask if the band itself, a 95% CI band for the mean response (variable on y-axis) as a function of the explanatory variable (variable on x-axis), is answering the question we are interested in asking.

Argument for including bands by default:

There is some mechanism we can invoke to convince ourselves our data is subject to random variation.  That variation means we should be somewhat uncertain about where the true line is, and some indication of the uncertainty, even if inappropriate, is better than no indication of uncertainty.

I don't have any problems with this exact example, more with assuming the band is meaningful in all cases.

novice-goals.Rmd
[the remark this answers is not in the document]
cwickham: I think it is useful to point out many different plots can be created using the same data and mapping just by swapping out geometric objects, and as part of that, it can be useful to see what running just e.g. `ggplot2::ggplot(hra, aes(x = log10.mass, y = log10.hra))` gives, i.e. a blank plot but with the x and y scales already set up.

But I think the plot object without data might be a bit too abstract.

novice-goals.Rmd
cwickham: I agree there should be a section on writing functions, and it should include a few assertions and checks.  What I'm not so sure about is where it should occur.

Let's assume I have a student who can string together enough code to get from raw data to a summary plot.  From the perspective of collaborating on a data analysis, it's more useful to me that they can put it all together with narrative in an Rmarkdown document, that it works both their and my computer, and they can get it on github, than it is that they can identify repetitive pieces and extract them into functions.

I guess I'm trying to say I think reproducibility should come before writing functions.

joelostblom: > Let's assume I have a student who can string together enough code to get from raw data to a summary plot. From the perspective of collaborating on a data analysis, it's more useful to me that they can put it all together with narrative in an Rmarkdown document, that it works both their and my computer, and they can get it on github, than it is that they can identify repetitive pieces and extract them into functions.

I very much agree with the passage above.

In addition, with learners who have never programmed before, I think it is crucial to focus on making them feel **comfortable** (in contrast to simply knowing how to do it) using programming instead of point and click. That is why I suggested repetition in a narrative format so that learners' familiarity with the key functions, packages, and analysis flow increases. I write this with the caveat that I mostly teach workshop style material and I can't speak to wether this approach is similarly effective or suitable in book format.

novice-goals.Rmd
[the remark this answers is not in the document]
gvwilson: Good points - added.

novice-goals.Rmd
gvwilson: Taken out Abela's Rules, added "which to choose" under plotting.

novice-goals.Rmd
gvwilson: Modified - please check.

novice-goals.Rmd
gvwilson: Some words added - please let me know.

novice-r-practice.Rmd
gvwilson: thanks - added

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   [Anscombe's Quartet and the DataSaurus dozen][anscome-datasaurus]
+-   What are the core elements of every plot?
+    -   data
+    -   geometric objects
+    -   aesthetic mapping
+-   How can I create different kinds of plots?
+    -   scatter plot
+    -   line plot
+    -   histogram
+    -   bar plot
```
lwjohnst86: At least for my field, it would be good to have a brief discussion on when *not* to use bar plots, since in biomedicine bar plots are frequently used inappropriately... but no one knows because its basically tradition to create them. I have some material on that to contribute.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   which to use when
+-   How can I plot multiple datasets at once?
+    -   grouping
+    -   faceting
+-   How can I make misleading plots?
+    -   showing a single central tendency data point instead of the individual observations
+    -   saturated plots instead of for example violins or 2D histograms
+    -   not starting axis at zero to exaggerate a difference
```
lwjohnst86: This I disagree with. Sometimes zero doesn't make sense, so why start at it. A better suggestion would "starting axis at a reasonable lower range". Take for instance body weight... it makes zero sense to include zero on the axis for body weight since that doesn't exist.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+-   How can I create different kinds of plots?
+    -   scatter plot
+    -   line plot
+    -   histogram
+    -   bar plot
+    -   which to use when
+-   How can I plot multiple datasets at once?
+    -   grouping
+    -   faceting
+-   How can I make misleading plots?
```
lwjohnst86: Regarding my comment above, this would be a good place to talk about using bar plots (with the mean and standard error), since they are incredibly misleading.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   showing a single central tendency data point instead of the individual observations
+    -   saturated plots instead of for example violins or 2D histograms
+    -   not starting axis at zero to exaggerate a difference
+    -   not using perceptually uniform colormaps to indicate quantities
```
lwjohnst86: This comment and the one below would be good to discuss e.g. viridis color scale.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+-   How can I represent and manage missing values?
+    -   NA
+-   How can I get a feel for my data?
+    -   summary statistics
+-   How can I create a simple model of my data?
+    -   formulas
+    -   linear regression
+        -   adding a best fit straight line on a scatterplot
+        -   understanding what the error bands on a "best fit" straight line mean
+    -   k-means cluster analysis
```
lwjohnst86: Any particular reason for showing k-means? Not sure how commonly used it is in other fields. I know most biomedical researchers don't ever use it. Maybe just discuss generally about clustering, mention a few methods, then focus on k-means?

novice-goals.Rmd
lwjohnst86: I definitely think if we are going to teach a segment on statistics, linear regression is the best choice since the concepts can be easily generalized to other methods, including non-regression based ones. But we also don't want to get too into statistics since this isn't a statistics course.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   diff
+    -   checkout
+-   How do I save work remotely?
+    -   push and pull
+-   How do I manage conflicts?
+    -   merge
+-   What *isn't* included?
+    -   forking
+    -   branching
+    -   pull requests
```
lwjohnst86: I definitely understand why this won't be included, but from my experience, learners "get it" when they get to collaborate together... but putting this into the course material is difficult and it is a difficult topic. So I'm conflicted on this.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   group
+    -   summarize
+    -   split-apply-combine
+-   How can I work with two or more datasets?
+    -   join
+-   How can I save my results?
+    -   writing files
+-   What *isn't* included?
+    -   anything other than reasonably tidy tabular data
+    -   map
```
lwjohnst86: I actually think that, while mapping as a topic is difficult, the implementation (at least in R) is relatively straight forward using purrr. Plus by at least showing it, the learners can see that this tool exists. And then they won't be tempted to using for loops (not good in R).

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   git reflow --substantive --single-afferent-cycle --ia-ia-rebase-fhtagn ...
+
+## Publishing
+
+-   How do static websites work?
+    -   URLs
+    -   servers
+    -   request/response cycle
+    -   pages
+-   How do I create a simple HTML page?
```
lwjohnst86: How in depth are we thinking about for this? Just enough to know what is happening when Markdown is converted to HTML?

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   head/body
+    -   basic elements
+    -   images
+    -   links
+    -   relative vs. absolute paths
+-   How can I create a simple website?
+    -   GitHub pages
+-   How can I give pages a standard appearance?
```
lwjohnst86: This could also be included with topics on R Markdown

novice-goals.Rmd
joelostblom: @lwjohnst86 We agree here. To clarify my sentence: "not starting an axis at zero **for the purpose of** exaggerating a difference". This doesn't preclude reasonable uses of non-zero axis limits, such as your example. For clarity, we can change the sentence accordingly:

```suggestion
    -   picking unreasonable axes limits to intentionally misrepresent the underlying data
```

novice-goals.Rmd
joelostblom: Yes, [the matplotlib docs](https://matplotlib.org/users/colormaps.html) has great info that can be simplified and abbreviated.

novice-goals.Rmd
joelostblom: I think kmeans is a good algorithm to introduce early because it is quite simple to understand the basic concepts of how it works. It also often performs well enough to at least be a good starting point when doing clustering/partitioning, even on high dimensional data. I have actually seen it used in many papers in my corner of the biomedical research field.

novice-goals.Rmd
cwickham: Yes! E.g [Dynamite plots](http://biostat.mc.vanderbilt.edu/wiki/Main/DynamitePlots). I think I was thinking barplots just as the categorical variable equivalent of a histogram. We should definitely emphasize appropriate and inappropriate uses.

novice-r-practice.Rmd
cwickham: See https://github.com/gvwilson/merely-useful.github.io/pull/2

gvwilson: thanks - merged.

novice-goals.Rmd
gvwilson: happy to have it, but I think this is getting out of outline and into content.

novice-goals.Rmd
gvwilson: What @joelostblom said: k-means is (in my experience) the simplest kind of clustering to teach.

novice-goals.Rmd
gvwilson: They can learn to collaborate by pushing to `master` in a shared repo - that's how we taught Subversion for years. Their cycle becomes `pull`-`merge`-`push`.

novice-goals.Rmd
gvwilson: I will defer to the R crew - @cwickham @joelostblom ? Do you think learners will get far enough to need `map` given everything else on the table?

novice-goals.Rmd
gvwilson: Yup.

novice-goals.Rmd
joelostblom: While I'm flattered to be included in the R crew, @lwjohnst86 is much more knowledgeable here and can answer this better than I can.

joelostblom:
@gvwilson I'm away over the weekend so I will get with responses to your comments and new comments in the latest version on Monday!

novice-goals.Rmd
gvwilson: :+1: - wherever it fits in most easily.

novice-r-practice.Rmd
```diff
@@ -0,0 +1,375 @@
...
+## Creating charts
+
+We will use data on the mass and home range area (HRA) of various species from:
+
+> Tamburello N, Côté IM, Dulvy NK (2015) Data from: Energy and the scaling of animal space use. Dryad Digital Repository.
+> https://doi.org/10.5061/dryad.q5j65
+
+```{r}
+hra <- readr::read_csv(here::here("novice-r-practice", "home-range-database.csv"))
+head(hra)
```
lwjohnst86: Given that `read_csv()` reads in the data as a `tibble()`, you don't need to use `head()` tibbles automatically only output the first 6 or so rows.

lwjohnst86:
```suggestion
hra
```

novice-r-practice.Rmd
```diff
@@ -0,0 +1,375 @@
...
+
+```{r chart-mass}
+ggplot2::ggplot(hra) +
+  ggplot2::geom_histogram(mapping = aes(x = mean.mass.g))
+```
+
+Try again with `log10.mass`:
+
+```{r chart-log-mass}
+ggplot2::ggplot(hra) +
```
lwjohnst86: In general, the `aes()` function is added here at the `ggplot(data, aes(...` phase, not at each individual `geom_`.

novice-r-practice.Rmd
```diff
@@ -0,0 +1,375 @@
...
+```
+
+*Create histograms showing the distribution of home range area using linear and log scales.*
+
+2. Change the visual appearance of a chart:
+
+```{r change-visual}
+ggplot2::ggplot(hra) +
+  ggplot2::geom_histogram(mapping = aes(x = log10.mass), bins = 100) +
+  ggplot2::ggtitle("Frequency of Species Masses") + ggplot2::xlab("Log10 of Mass") + ggplot2::ylab("Number of Species") +
```
lwjohnst86:
```suggestion
  ggplot2::ggtitle("Frequency of Species Masses") + 
  ggplot2::xlab("Log10 of Mass") + 
  ggplot2::ylab("Number of Species") +
```

lwjohnst86: Newline after each `+` as per the tidyverse coding style. https://style.tidyverse.org/

novice-r-practice.Rmd
```diff
@@ -0,0 +1,375 @@
...
+*Group by order and experiment with different alpha values.*
+
+5. Create a faceted plot:
+
+```{r facet-plot}
+hra %>%
+  dplyr::mutate(class_fct = as.factor(class)) %>%
+  ggplot2::ggplot() +
+  ggplot2::geom_point(mapping = aes(x = log10.mass, y = log10.hra, color = class_fct), alpha = 0.5) +
+  ggplot2::facet_wrap(~class_fct)
```
lwjohnst86:
```suggestion
  ggplot2::facet_wrap(vars(class_fct))
```

lwjohnst86: As per the latest ggplot2, variables given to `facet_wrap()` should be wrapped in the `vars()` function.

novice-r-practice.Rmd
```diff
@@ -0,0 +1,375 @@
...
+*Create a plot faceted by order for just the reptiles.*
+
+6. Fit a linear regression to the logarithmic data for birds:
+
+```{r fit-line}
+hra %>%
+  dplyr::filter(class == "aves") %>%
+  ggplot2::ggplot() +
+  ggplot2::geom_point(mapping = aes(x = log10.mass, y = log10.hra), alpha = 0.5) +
+  ggplot2::geom_smooth(method = lm, mapping = aes(x = log10.mass, y = log10.hra), color = 'red', se = FALSE)
```
lwjohnst86: Putting the `aes()` in `ggplot()` would reduce the amount of `aes()` calls in each `geom_`

novice-goals.Rmd
gvwilson: Can't create a pull request within a single branch - has to be between branches or between repos.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+    -   frame these as exploratory tools for revealing structure in the data, rather than modelling or inferential tools
+-   How can I put people at risk?
+    -   algorithmic bias
+    -   de-anonymization
+-   What *isn't* included?
+    -   statistical tests
+    -   multiple linear regression
+    -   anything with "machine learning" in its name
+
+## Version control
```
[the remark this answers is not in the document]
gvwilson: Depends how much material we want to share between the novice R and novice Python stuff - if we use Git Kraken, we can use a single lesson for both, but if we use the IDE(s), we'll have to do separate lesson materials.

novice-goals.Rmd
```diff
@@ -0,0 +1,226 @@
...
+
+-   How can I make programs easy to read?
+    -   coding style
+    -   linters
+    -   documentation
+-   How can I make programs easy to re-use?
+    -   Taschuk's Rules
+-   How can I combine explanations, code, and results?
+    -   notebooks
+-   How should I organize my projects?
```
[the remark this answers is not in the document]
gvwilson: added

novice-r-practice.Rmd
```diff
@@ -0,0 +1,375 @@
+# Practice
+
+We have covered a lot in the last few lessons,
+so this one presents some practice exercises to ground what we have learned
+and introduce a few more commonly-used functions.
+
+## Working with a single tidy table
+
+1. Load the tidyverse collection of libraries and the `here` library for constructing paths:
```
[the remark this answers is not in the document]
gvwilson: fixed.

novice-goals.Rmd
lwjohnst86: Mm, I misunderstood your earlier text. Makes sense.

novice-r-practice.Rmd
gvwilson: Garrett Grolemund's examples add aesthetics inside geoms (at least, some of them do), and that was the model I was following. Happy to change if the R crowd thinks it should be.

novice-r-practice.Rmd
gvwilson: fixed.

novice-r-practice.Rmd
gvwilson: fixed

novice-r-practice.Rmd
gvwilson: fixed - thanks.

novice-r-practice.Rmd
cwickham: I agree with Luke here.

gvwilson: Sold!

novice-goals.Rmd
```diff
@@ -0,0 +1,227 @@
+# Outline {#outline}
+
+This outline describes the questions that the novice courses on R and Python will answer.
+The advanced course can then assume that learners have hands-on experience with these topics
+but nothing more.
+
+## Getting started
```
joelostblom: @gvwilson I would put the brief intro to computation as the first section and it would include the following:

-   Humans communicate with computers by modulating flows of current to instruct the computer to carry out tasks that are to complex to do by hand.
-   Transition from moving physical wires to graphical point and click interfaces.
-   Programming languages are formal languages, just like mathematics. This makes them effective and with less ambiguity than natural languages.
-   Communicating via typing text might seem more primitive than pint and click graphics, but has many advantages. Compare to using a dictionary to learn a language.
-  Conceptual understanding example of graphical vs text: Instead of using the mouse to click the "File" menu, then on "Open file", and then find the filename and click it, we would type `file.open('filename')`.

I am afraid this is a bit wordier than the bullets in the rest of the outline, I just wanted to ensure it is clear. I can draft the first version of this when we get into writing the content.

novice-goals.Rmd
```diff
@@ -0,0 +1,227 @@
...
+## Development
+
+-   How can I make my own functions?
+    -   declaring functions
+    -   declaring parameters
+    -   default values
+    -   common conventions
+-   How can I make my programs tell me that something has gone wrong?
+    -   validation (did we build the right thing) vs. verification (did we build the thing correctly)
+    -   assertions for sanity checks
```
joelostblom: @gvwilson For exploratory data analysis, I am also in favor for understanding how to carry out less automatic sanity checks such as plotting to check that a data transformation makes sense or seeing if there are extreme values impacting your data summaries (or feature scaling). Similar to what @cwickham mentioned earlier:

> Testing in this context might be less formal than unit tests and involve data summaries, tables or plots that the reader/analyst can use to verify things went as planned.

DamienIrving:
Passed at meeting on 29 May 2019.

PR #80, opened 2019-05-28
Moving meetings to wiki

gvwilson:
I've created wiki pages with legacy content - would appreciate a check. (In particular, did we actually have a meeting on May 9, or is that file a placeholder for the meeting that Madeleine moderated? If so, please delete the 2019-05-09 wiki page.)


mbonsma:
@gvwilson there was a meeting - it was the Python tool choice meeting. It was on May 9, not May 8 - I've updated the page date accordingly in the wiki.

PR #127, opened 2019-06-18
Add missing values materials, first

brandeism:
Missing Data section: A few learning objectives, lecture notes and exercises added

### DIFF
--- a/novice-goals.Rmd
+++ b/novice-goals.Rmd
@@ -233,14 +233,99 @@ but nothing more.
     -   Calculate the average of column XXX
 
 -   How can I create a simple model of my data?
+    Python Learning Objectives 
+    -   Describe clustering and regression
+    -   Define k-means clustering and linear regression analyses
+    -   Demonstrate appropriate applications or uses of k-means clustering analysis
+    -   Demonstrate appropriate applications or uses of linear regression analysis
+    -   State and specify limitations of both analyses
+
+    Lecture Notes (K-means Cluster Analysis)
+    -   Resource: Information to Information Retrieval (https://nlp.stanford.edu/IR-book/information-retrieval-book.html), Chapter 16.
+    -   Clustering: the process of grouping a set of elemens into classes of similar elements
+    -   Clustering is also called unsupervised learning, e.g., "learning" from raw data
+    -   K-means clustering refines goups iteratively where each element belongs to only one group 
+        (a nod to flat algorithm approaches w/ hard clustering practices, if we want to expand in later books)
+    -   Algorithm setup
+        -   Assumes elementss are real-valued vectors
+        -   Clusters based on centroids (or mean) of points in a cluster
+        -   Reassignment of elements to clusters is based on distance to the current cluster centroids.
+    -   Algorithm Pseudocode
+        -   Step 0: Select K random elements {s1, s2,...,sK} as seeds. These are your initial K centroids.
+        -   Step 1: Assign points to closest centroid
+        -   Step 2: Recompute cluster centroids
+        -   Step 3: Goto Step 1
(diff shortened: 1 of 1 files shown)


novice-goals.Rmd
```diff
@@ -102,6 +102,46 @@ but nothing more.
...
+    -   Many descriptive statistics methods have skipna option to control if missing data should be excluded . This value is set to True by default (unlike R) 
+
+    Python Lecture Notes (useful commands)
+    -   Table: df.method() / description
+    -   dropna() / drop missing observations
+    -   dropna(how='all') / drop observations where all cells are NA
+    -   dropna(axis=1,how='all') / drop coloumn if all values are missing
+    -   dropna(thresh=5) / drop rows that contain less than 5 non-missing values
+    -   fillna(0) / replace missing values with zeros
+    -   isnull() returns True if the value is missing
```
joelostblom: For consistency with `fillna()` and `dropna()`, I suggest we use the alias `isna()` instead.

novice-goals.Rmd
```diff
@@ -102,6 +102,46 @@ but nothing more.
...
+    -   isnotnull() / Returns True for non-missing values
```
joelostblom: This is without the "is"-part, and similarly to above, I suggest we use `notna()` instead.

novice-goals.Rmd
```diff
@@ -102,6 +102,46 @@ but nothing more.
...
+
+    T/F Exercises (#Review4Everybody)
+    -   Imputation is really just making up data to artificially inflate results.  It’s better to just drop cases with missing data than to impute.
```
joelostblom: I think there are some situations, such as timeseries, where imputations are still handy, especially when observations need to have the same periodicity and can't contains NaNs for downstream analysis. I agree with you below that a brief stat intro might be needed to fully grasp these concepts.

novice-goals.Rmd
```diff
@@ -102,6 +102,46 @@ but nothing more.
...
+    Python Lecture Notes (general info) 
+    -   Missing Completely at Random (MCAR)
+    -   Missing at Random (MAR)
+    -   Missing Not at Random (MNAR)
+    -   Missing data at scale (what's a little and what's alot?)
+    -   Imputation and Multiple Imputation 
+    -   When summing the data, missing values will be treated as zero
+    -   If all values are missing, the sum will be equal to NaN
+    -   cumsum() and cumprod() methods ignore missing values but preserve them in the resulting arrays
+    -   Missing values in GroupBy method are excluded (just like in R)
```
joelostblom: Potentially worth mentioning that using `.groupby()` with [`observed=True` includes NaNs for categoricals](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.23.0.html#new-observed-keyword-for-excluding-unobserved-categories-in-groupby).

novice-goals.Rmd
brandeism: Sounds good, @joelostblom. I'll revise

novice-goals.Rmd
brandeism: Yeah, I contemplated adding .groupby() too. If it's not included elsewhere, it'll be added here.

ChristinaLK:
No specific comments except to agree w/ the discussion that the stats intro (or at least part of it?) should go before the missing data section for context *and* I like the separation of general principles from python how-tos -- it makes sense that you provide the context for what you're doing (this and that about missing values) and then how to actually put it into practice using Python tools.

gvwilson:
What's the status of this PR please?

gvwilson:
Is this still live or has it gone stale?

PR #73, opened 2019-05-23
WIP: Bibliography issues

ljdursi:
First stab at #16 to improve bibliography.  Explicitly uses a common CSL file (for J Cell Biology) in both pdf and html outputs, which includes `note` fields

This PR will be complete when it addresses the three points listed in the issue:

- [ ] Single bibliography page in HTML format - this seems to be an RMarkdown thing
- [ ] Use a brief reference format 
- [x] Include notes field (although formatting improvements might be useful)

PR #131, opened 2019-06-25
Bibliography fixes

### LINKED ISSUE #16: Better formatting for bibliography
1. Single bibliography page instead of citations at the end of each chapter (which will become more important as the book grows and we add reviews or notes to bibliography entries).
2. Use a brief reference format rather than the length default of `apalike`.
3. Include `note` entries in bibliography.

lwjohnst86:
Resolves #16. From #73 and will close it.

Looks like this:

Screenshot from 2019-06-25 23-07-11

@gvwilson is that what you were thinking?


gvwilson:
:+1: to merge once the conflicts are resolved - thank you.

Issue #96, opened 2019-06-01
Novice R section assignment: Data Manipulation


cwickham:
This issue tracks the [Data Manipulation](https://github.com/merely-useful/merely-useful.github.io/blob/925ff17ea8b0b5aa5fa15d219a8b9012e533397c/novice-goals.Rmd#L22-L49) section for the novice R material.

This issue can be closed when:

* [ ] There are an initial set of R exercises corresponding to the outline
* [ ] There is a first draft of the text of the section

lwjohnst86:
Since the branch was merged, closing this.

PR #137, opened 2019-06-27
Novice R Data Manipulation

cwickham:
Will eventually address #96


cwickham:
A little rearranging compared to outline:

* Moved data import to after all other manipulation - assuming the dog licenses data will be in R package at some point so don't need to talk about data import until later.
* `arrange()` is covered first outside the question "How do I select subsets of the data?" as the first example of a dplyr verb.

Some things I think are still missing:
* the whole `join()` section is missing
* introduction to the data (maybe this happens in this chapter, maybe earlier?)
* Need to introduce `count()`
* `reprex::reprex()`? 
* assorted `TODO`s throughout text

gvwilson:
thanks - looking forward to it.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+This join can be performed with the `left_join()` function in dplyr:
+```{r}
+dogs %>% left_join(neighborhoods)
+```
+It's called a `left_join()` because we will get (at least) one row back for every row in the dataset on the left, in this case, `dogs`.  I like to think about `left_join()` working just as we described above: for each row in the dataset on the left, look up the value in the dataset on the right. 
+
+Notice that `left_join()` gave us the message `Joining, by = "nta"`.  The join functions in dplyr will use any columns with the same name in both datasets to do the join.  If the datasets have columns for the same quantity but with different names, for instance, if `neighborhoods` had a column called `nta_code` instead of `nta`:
+```{r}
+neighborhoods_2 <- neighborhoods %>% 
+  rename(nta_code = nta)
```
lwjohnst86: rename hasn't been covered yet, so it should be mentioned earlier on.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+```
+
+Then attempting the join will result in an error - `data sources have no common variables`:
+```{r, error = TRUE}
+dogs %>% left_join(neighborhoods_2)
+```
+You can use the `by` argument to the join function to specify which columns are to be used to do the join:
+```{r}
+dogs %>% 
+  left_join(neighborhoods_2, by = c("nta" = "nta_code"))
```
lwjohnst86: Might need to explain what the code for the `by` argument is doing and how to use.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+```
+
+### Exercise
```
lwjohnst86:
```suggestion
### Exercise: Predicting the output of a join
```

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+Try to predict what this code will return, then run it and check your prediction:
+```{r, results = "hide", message = FALSE}
+neighborhoods %>% left_join(dogs)
+```
+
+### Exercise: Dog licenses per capita
+
+* Use `left_join()` to join `dogs_by_nta` with `nta_pop`. 
+
+* `filter()` the result to just use population from `year` 2010.
```
lwjohnst86:
```suggestion
* `filter()` the result to only use population from `year` 2010.
```

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+  left_join(nta_pop, by = c("neighborhood_tabulation_area" = "nta_code")) %>% 
+  mutate(licenses_per_person = n_licences/population) %>% 
+  filter(year == 2010) %>% 
+  arrange(desc(licenses_per_person)) %>% 
+  select(nta_name, licenses_per_person)
+# Stuyvesant Town-Cooper Village averages 0.045 dogs per person
+# Or ~45 dogs per 1000 people
+```
+
+### Other join functions
```
lwjohnst86: How important is it to describe these other joins? I personally don't use many other than `left_join` and `full_join` and often struggle with the logic of the other joins. Would it be better to simply say "there are other joins out there, check out the cheatsheet/linked tutorial if you want to learn more, but we won't cover them more here". Thoughts?

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+```
+
+
+### Exercise
+
+Find the dog licences per capita, but now at the community district level, by joining the `dog_licenses` data with the `nyc-cd-pop` data.
+
+## How can I read my own tabular data into R?
+
+### What is tabular data?
```
lwjohnst86: This should be the first lesson of the chapter.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+Here's how the table above would look inside the CSV file `site.csv`:
+```{r, echo = FALSE, comment = NA}
+writeLines(readLines(example_csv))
+```
+
+In this case the first line has the column names, this is common (and recommended!), but not universal. 
+
+Why is CSV so popular? 
+
+* It's human readable.  CSV isn't a special file type, it is just a plain text file that follows some conventions.  This means you don't need any special software to look at the contents -- you can open in up in anything that can examine text and take a look inside.
```
lwjohnst86:
```suggestion
* It's human readable.  CSV isn't a special file type, it is a simple plain text file that follows some conventions.  This means you don't need any special software to look at the contents -- you can open it up in anything that can examine text and take a look inside. 
```

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+```{r, echo = FALSE, comment = NA}
+writeLines(readLines(example_csv))
+```
+
+In this case the first line has the column names, this is common (and recommended!), but not universal. 
+
+Why is CSV so popular? 
+
+* It's human readable.  CSV isn't a special file type, it is just a plain text file that follows some conventions.  This means you don't need any special software to look at the contents -- you can open in up in anything that can examine text and take a look inside. 
+* It's computer readable.  Because CSV files all have the same structure it's easy to write computer programs to read them.  This means in almost any program designed to work with data, and all the common programming languages, you'll find functions that will import CSV files.  This also means it's easy to create CSV files -- you can export them from Excel, write them from R, or even write one from scratch in a text editor.
```
lwjohnst86:
```suggestion
* It's computer readable.  Because CSV files all have the same structure it's easy to write computer programs to read them.  This means in almost any program designed to work with data, which is basically all the common programming languages, you'll find functions that will import CSV files.  This also means it's easy to create CSV files -- you can export them from Excel, write them from R, or even write one from scratch in a text editor.
```

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+### Importing CSV data into R
+
+To work with data in R you need to have it in R's memory.  The `read_csv()` function in the readr package will import a CSV file, and represent it as a tibble, if you give it the location of the CSV file.  For example, to read the `site.csv` data and store it in an object called `sites`: 
+```{r}
+library(tidyverse)
+sites <- read_csv(here("data", "site.csv"))
+sites
+```
+
+TODO: talk about files paths, point reader to the place where file paths are talked about, or assume file is in their working directory.
```
lwjohnst86: File path discussion also happens in the reproducibility section, so not sure how we want to coordinate that.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+
+Notice that `read_csv()` gave us a message about what it did: it parsed our data file and found three columns `site_id`, `latitude` and `longitude`.  It also mentions what kind of data it assumed was in each column.  TODO: point to further discussion of data types.
+
+The object `sites` is now R's representation of the data from the `site.csv` file.    
+
+### Exercise: Import `visited.csv`
+
+Use `read_csv()` to read `visited.csv` into R. How does R indicate a cell with a missing value?
```
lwjohnst86: I assume they'll have access to this dataset? How do we get them to download it, etc?

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+library(here)
+dog_licenses <- read_csv(here("data", "nyc-dog-licenses.csv.gz")) %>% 
+  drop_na(animal_name)
+```
+
+TODO: where is data introduced?
+
+TODO: clean version with no missing values and snake case column names, including month and year columns for some (all?) of the date columns.
+
+TODO: I'm assuming that at some point this data will live in R package, so we can delay importing data to end of chapter.
```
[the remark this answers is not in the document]
cwickham: Yeah, maybe...

Pros: 
- we can get students working with data very quickly without digressing into paths, CSVs etc.
- we don't have to worry about getting the data files to the students
- we can have documentation available for the data in the students' R session

Cons: 
- students need to install this data package (will it be on CRAN (more work for us), or will this involve `remotes::install_github()` (more work for students)?)
- doesn't reflect real-life data analysis
- delays discussion of import _too_ long?

I'm mostly thinking about this for the early chapters.  I certainly think we need to talk about import, and I would want students importing their own data by the end of the course, but early on, having data in a package might get students doing fun things quicker.

Some of the pros could equally be handled on the admin side in a classroom situation by, for example, using rstudio.cloud with a project with the data files included.   I think that is a fine alternative...

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+
+The RStudio Viewer has an interface much like other spreadsheet programs you might have used.  You can use this Viewer to look at the `dog_licenses` tibble with the `View()` function:
```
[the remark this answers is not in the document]
cwickham: Yes, this needs to be somewhere.  If the data is in a package this becomes a one liner ;)

r-data-manipulation.Rmd
cwickham: Good catch!  This is one of my bad habits...

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+  select(animal_name, is_chase, called_chase)
+```
+
+Arithmetic is another common operation that returns as many elements as there are rows.  For instance we could see how long licenses are issued for:
+```{r}
+dog_licenses %>% 
+  mutate(license_duration = license_expired_date - license_issued_date) %>% 
+  select(license_duration)
+```
+Interesting that there is so much variation here.
```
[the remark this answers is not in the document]
cwickham: Yeah, this is more a personal observation.  I was expecting licenses to be issued for standard periods like one year.  I think either I expand this statement to make this observation more clearly, drop it entirely, or have a more complete exploration of it somewhere else - it seems ripe for a histogram.  

For now, how about:
```suggestion
Looks like licenses aren't issued for a standard time period.  In these first rows, there are some licenses issued for a whole number of years: 2 (731 days), 3 (1097 days) and 5 (1826 days).  However, others seem to be for fractions of years like 421 days.
```

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+
+The function `str_length()` in the stringr package finds the length of character strings.  Replace the `___` in following code to add a column called `name_length` that contains the length of the dog's name:
+```{r, eval = FALSE}
+dog_licenses %>% 
+  ___(___ = stringr::str_length(animal_name))
+```
+
+Now add an `arrange()` step to find the licenses issued to dogs with the longest names?
+
+## How can I tell what's gone wrong in my programs?
```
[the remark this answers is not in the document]
cwickham: I agree it feels a little out of place.  

It does seem like a discussion that should happen before too much is covered, but it doesn't have to be here. If I was teaching in a semester course I feel like I'd cover this at the end of lecture 1 or start of lecture 2...maybe not in its entirety, but at least to say "things will go wrong, and that's OK, here's some tips on getting things going again".

Its position in the book probably doesn't matter to much for an instructor-led class since they can jump to it when they feel it's appropriate (as long as the examples don't assume too much).  Where would it be most useful, and least distracting, for someone working through the book on their own?

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+      filter(animal_name == BRUNO)
+    ```
+
+### Warnings and messages 
+
+There are two other kinds of alerts R can give: warnings and messages.  These can both appear in the console with the same color as an error, but they are informational as opposed to fatal.  
+
+Messages are purely informational, for example when you read data in with `read_csv()` you get a message that describes the columns and their data types as parsed by the function:
+```{r}
+sites <- read_csv(here("data", "site.csv"))
```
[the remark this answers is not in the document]
cwickham: I think I prefer this second option: avoid `here()` and `"data"`.   I'll make that change now, and we may need to revisit depending on where this section ends up.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+```
+the `filter()` step came after the `mutate()` step. Does this matter?  
+
+* Swap the order in the code and see if you get the same results.
+
+* Write out how you might describe the steps. Is it obvious you can swap the filter and mutate step and get the same results?
+
+* Which steps can't you swap the order of? Why?
+
+* *There are some performance differences...can you guess why?*
```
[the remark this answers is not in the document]
cwickham: Reworded to:
Despite giving the same results, some orderings of the data manipulation steps will take longer to compute. Can you guess why?

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+```{r}
+nta_pop %>% 
+  arrange(desc(population)) %>% 
+  select(nta_code, nta_name, year, population, year)
+```
+You'll see that MN12, the Upper West Side, in fact has the largest population of all the neighborhoods (at least in 2000 and 2010), so perhaps it isn't surprising it also has the most dog licenses issued in 2016.
+
+Joins are a way to combine two datasets together by using a column the datasets have in common - like the columns that contain the NTA codes.  Let's work with some smaller datasets to see how this works.  
+```{r, include = FALSE}
+dogs <- tribble(
```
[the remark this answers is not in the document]
cwickham: This code chunk isn't in the text.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+  ~animal_name,  ~nta, ~breed_name,
+  "PRECIOUS",  "QN12",  "Yorkshire Terrier",
+  "TOBY",      "QN55",  "Jack Russell Terrier",
+  "GUCCI",     "QN01",  "Yorkshire Terrier",
+  "HUNTER",    "QN12",  "Unknown", 
+)
+neighborhoods <- nta_pop %>% 
+  filter(nta_code %in% c("QN01", "QN02", "QN03", "QN12"), year == 2010) %>% 
+  select(nta = nta_code, population, nta_name)
```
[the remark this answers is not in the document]
cwickham: Same here, this code chunk isn't in the text.

r-data-manipulation.Rmd
cwickham: I think I'd rather avoid it, to not distract from the explanation here.  I'll rework to just show the data, rather than show how it was created.

r-data-manipulation.Rmd
cwickham: I almost only use `left_join()`.  So, probably not very important...

r-data-manipulation.Rmd
cwickham: I can appreciate following the logical order of a data analysis project, e.g. import first, but I worry leading with import doesn't build enough excitement.  I.e. why would I want to get my data into R, if I'm not entirely convinced I'll be able to do useful things once I do?  

I have some similar concerns with leading with data manipulation, rather than say plotting.  

Can we table the ordering discussion until we have some of the other sections drafted?

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1111 @@
...
+arrange(name_and_breed, breed_name)
+```
+
+### Filtering to keep a subset of rows
+
+The function to filter rows of a data frame is `filter()`.  Like `arrange()` and `select()`, its first argument is a dataframe or tibble.  The remaining arguments describe which rows to **keep**.  The rows to keep are specified with a logical expression - something that is either `TRUE` or `FALSE`. The rows where this expression is `TRUE` will be returned.
+
+One of the simplest kinds of logical expression is a test for equality with the `==` operator.  For example, to keep the rows where `animal_name` is `BRUNO` you could do:
+```{r}
+dog_licenses %>% filter(animal_name == "BRUNO")
```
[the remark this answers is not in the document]
cwickham: Added this.

r-data-manipulation.Rmd
cwickham: Good question.  Do we:

* link to it, and require a student to download to an appropriate location, 
* give code to download it to an appropriate location, 
* have the raw data as part of a package that a student has installed and give code to get its path 
* rely on the "instructor" to provide it (i.e. in a rstudio cloud project)?

Other ideas? Thoughts?

r-data-manipulation.Rmd
cwickham: I'm leaning towards assuming the file is in their working directory (and I guess having to talk about working directories, or projects) and pointing to reproducibility section for more discussion.

cwickham:
@gvwilson this has had the once over from Luke.  Now would be a great time to give it a look.  Would also appreciate you weighing in on some of our unresolved conversations.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+    ```{r}
+    dogs %>% left_join(neighborhoods)
+    ```
+
+* `right_join()`, is just like `left_join()` but with the roles of the data sets reversed.
+    ```{r}
+    dogs %>% right_join(neighborhoods)
+    ```
+
+* `inner_join()` returns only rows that have matches.
```
gvwilson: We're going to have to test-drive this with a few learners - again, it feels like a leap in complexity.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+    ```{r}
+    dogs %>% inner_join(neighborhoods)
+    ```
+
+* `full_join()` returns all rows from both datasets, using missing values in columns where a row doesn't have a match in the other dataset.
+    ```{r}
+    dogs %>% full_join(neighborhoods)
+    ```
+
+Another useful join is `anti_join()`.  It doesn't actually return a combined dataset, you only get columns from one dataset back, but it returns the rows in the left hand dataset with no matches in the right hand dataset.  This is a great way to explore what you might be missing.  For example, it makes it easy to ask which rows in `dogs` don't have matches in `neighborhoods`:
```
gvwilson: does `anti_join` also returns rows from right with no match from left?

r-data-manipulation.Rmd
gvwilson: :+1: we'll fix in the edit

r-data-manipulation.Rmd
gvwilson: I think we package it...

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+  filter(borough == "Queens") %>% 
+  summarise(max(animal_birth_month))
+```
+You haven't learnt about the `filter()` or `summarise()` functions yet, but can you guess what the code does? What question might it answer? 
+
+### Exercise:  Using the pipe
+
+Re-write this snippet of code to use the pipe:
+```{r, results = "hide"}
+select(dog_licenses, animal_name, breed_name)
```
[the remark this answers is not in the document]
cwickham: Yes.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+or you've forgotten that you've used a separator
+```{r, error = TRUE}
+myobject
+```
+
+This error also often arises when you forgot quotes around strings.  For example, if we want all the dogs that are male, and try
+```{r, error = TRUE}
+dog_licenses %>% filter(animal_gender == M)
+```
+you get an error because R is looking for an object called `M` to compare to the values in the column called `animal_gender`.  What we really wanted to do was compare the values in `animal_gender` to the string `"M"`:
```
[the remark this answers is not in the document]
joelostblom: What happens when there is a variable and a column with the same name?

r-data-manipulation.Rmd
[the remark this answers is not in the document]
joelostblom: I like the URL option, I made a comment in the issue suggesting this instead of making a package.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+bad
+```
+
+Warnings don't **stop** you from proceeding, but they should alert you to question whether you **should be** proceeding.
+
+### What do I do when I get an `Error` I can't fix?
+
+* Check that the error is reproducible.  Restart R with a clean slate and re-run your code up to and including the code that gives the error. This is the R version of the classic tech advice to "turn it off, then turn it on again".  TODO: link to reproducibility chapter.
+
+* Try googling it: for example, googling "R unexpected string constant" lead me to the question ["Error: unexpected symbol/input/string constant/numeric constant/SPECIAL in my code" on StackOverflow][error-unexpected] which gives some great examples of ways this error might arise.
```
joelostblom: Can we be search engine neutral and say "Try searching for it online" or similar instead? Or does that make us sound like dinosaurs?

(On a more personal note, I like duckduckgo, but I have already mentioned "rubber duck debugging" and "ducktyping" in my python lecture, if I add "duckduckgoing" instead of "searching online", questions will be asked.) =)

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+* Ask for help. You are most likely to get help when you can provide a reproducible example. TODO: provide link
```
joelostblom: I wrote about this in the py-dev lessons in case you want to have something similar here: https://github.com/merely-useful/merely-useful.github.io/blob/4b2cdc726da061908435e7ebc867ed21b258b5e0/py-development.Rmd#L504-L516.

links
[so-mre]: https://stackoverflow.com/help/minimal-reproducible-example
[so-mre-r]: https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+Tabular data describes data that is in the form of a table: values arranged in rows each of the same length, or equivalently values arranged in columns each of the same length. Here's a small example of some tabular data:
+```{r, echo = FALSE, message=FALSE}
+example_csv <- here("data", "site.csv")
+sites <- read_csv(example_csv )
+knitr::kable(sites)
+```
+
+Each row records information on a site at which water measurements are taken.  There are three columns: a site identification code, and the location of the site in latitude and longitude. 
+
+This is an incredibly common way of *displaying* data, but when *storing* tabular data in a file, we need a way to communicate when records and values begin and end.  A very popular format for doing this is CSV.  CSV, is short for **c**omma **s**eparated **v**alues, and like the name suggests, a comma, `,`, is used to separate the values for each column, while each record goes on a new line.  The file is plain text, but we use the extension `.csv` to indicate that is follows the CSV format conventions.
```
joelostblom: I think this explanation plus 1-2 sentences on that a csv file can be stored on the internet and accessible via a url is enough for student to understands what happens when reading in a csv from a url.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+This data includes populations for each NTA and a full name for the NTA.  If you look at the neighborhoods with the largest populations:
+```{r}
+nta_pop %>% 
+  arrange(desc(population)) %>% 
+  select(nta_code, nta_name, year, population, year)
+```
+You'll see that MN12, the Upper West Side, in fact has the largest population of all the neighborhoods (at least in 2000 and 2010), so perhaps it isn't surprising it also has the most dog licenses issued in 2016.
+
+Joins are a way to combine two datasets together by using a column the datasets have in common - like the columns that contain the NTA codes.  Let's work with some smaller datasets to see how this works.
```
joelostblom: I would be inclined to introduce append/concat functionality (I think `bind_row`/`bind_col` in dply?) before joins since these are simply adding rows or columns which I think is simpler to understand than adding based on the values in a common column.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+### Exercise: Reading aloud
+
+Read the following code aloud to your neighbor (or cat, dog or rubber duck).  Remember to pronounce `%>%` as "and then".  
+```{r, eval = FALSE}
+dog_licenses %>% 
+  filter(borough == "Queens") %>% 
+  summarise(max(animal_birth_month))
+```
+You haven't learnt about the `filter()` or `summarise()` functions yet, but can you guess what the code does? What question might it answer?
```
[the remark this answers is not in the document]
cwickham: Yeah, fair point.  I'll rewrite to use functions seen.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+Use the pipe to re-write this snippet of code to avoid the intermediate variable:
+```{r, results = "hide"}
+name_and_breed <- select(dog_licenses, animal_name, breed_name)
+arrange(name_and_breed, breed_name)
+```
+
+### Filtering to keep a subset of rows
+
+The function to filter rows of a data frame is `filter()`.  Like `arrange()` and `select()`, its first argument is a dataframe or tibble.  The remaining arguments describe which rows to **keep**.  The rows to keep are specified with a logical expression - something that is either `TRUE` or `FALSE`. The rows where this expression is `TRUE` will be returned.
```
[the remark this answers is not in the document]
cwickham: This is something I kept thinking about while writing and never really resolved. This resulted in a mix of me using tibble and data frame like synonyms in some places, and like distinct entities in others (like here).  

I think the resolution is probably to pick one term and use it exclusively, but have an explanation somewhere about what the difference is and our convention for using one term.

I'll go through and consistently use tibble.  Then at least it is a simple "search and replace" if there is a preference for data frame.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+The logical expression, `animal_name == "BRUNO"`, will be `TRUE` when the value of the `animal_name` column is exactly equal to the character string `"BRUNO"` - any differences in case, or whitespace will result in `FALSE`.
+
+Columns names can appear on either side of `==` or even on both sides.  When column names appear on both sides, the comparison happens row by row.  In programming terms this is know as an element-wise comparison.  For example, `license_issued_date == animal_birth_month` will return `TRUE` for a row only if for that row the date the license was issued is the exact same date as the birth month for the dog. If you take a look:
+```{r}
+dog_licenses %>% 
+  filter(license_issued_date == animal_birth_month)
+```
+There are no rows that satisfy this criteria.
+
+You can think of `==` as being a specific kind of comparison.  Other comparisons include:
```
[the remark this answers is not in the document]
cwickham: Yes, fixed.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+If you find yourself combining lots of comparisons on the same column with `|`, like dogs named `BRUNO`, `BRUCE` or `BRADY`:
+```{r, results = "hide"}
+dog_licenses %>% 
+  filter(animal_name == "BRUNO" | animal_name == "BRUCE" | animal_name == "BRADY")
+```
+You can save a lot of typing with `%in%`:
+```{r, results = "hide"}
+dog_licenses %>% 
+  filter(animal_name %in% c("BRUNO", "BRUCE", "BRADY"))
```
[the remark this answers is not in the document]
cwickham: Probably not...it seems to not be in Getting Started.  I'll add a little explanation here.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+endt_of_2016 <- as.Date("2016-12-31")
+```
+
+Use `filter()` with these variables to find the dog licenses that expire during  2016.
+
+## How can I calculate new values?
+
+So far, you've been manipulating the columns that already exist in a dataset, but what if you want to add new ones? The function `mutate()` handles this kind of operation.  
+
+To see how this works let's start with a logical expression: `animal_name == "CHASE"`.  If you used this with `filter()`, you would get all the rows back where the license was issued to a dog named `CHASE`. Let's say instead of subsetting the data, you want to add a column called `called_chase` that contained the `TRUE` and `FALSE` result.  With `mutate()` after passing in the data, you pass named arguments, where the name is the name you desire for the new column, and its value is the way to calculate it:
```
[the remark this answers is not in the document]
cwickham: Added an example

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+You generally want to make sure you've ruled out syntax errors before assuming it's a runtime error. Unfortunately, R doesn't distinguish these two types of error in its output, so we'll discuss some of the most common examples in the following sections.
+
+This analogy is also a reminder that sometimes you'll have to repeat yourself.  With R, giving the exact same instruction should always result in the exact same error, but it's not uncommon, even for longtime R users, to run and edit a line of code multiple times before it runs without error.
+
+There is one flaw in this analogy -- R isn't a real person. So, if you need to vent *your* frustration by cursing it, insulting its parentage or storming off, it's fine, no one's feelings will be hurt.  Of course, it won't change R's response...
+
+### Common syntax errors
+
+Syntax errors occur when your code can't be parsed into its component pieces by R.  For example, R expects the arguments to a function to start after the opening parenthesis (`(`), be separated by a comma (`,`) and finish at the closing parenthesis (`)`). When R sees this code:
```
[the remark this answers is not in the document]
cwickham: switched to "broken"

r-data-manipulation.Rmd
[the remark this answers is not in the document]
cwickham: Ugh yes, I know!  I tried [writing something](https://github.com/cwickham/quotation/blob/master/quotes.md) up once to help people with the "when do I need quotes?" question, but I'm not sure it was very useful.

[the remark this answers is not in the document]
cwickham: Good point Joel! Also, if you put quotes around a column name, you get no error, just not the result you were hoping for.  This feels like it needs it's own section... "where do quotes go in dplyr functions?"

r-data-manipulation.Rmd
cwickham: Good suggestion...I also use duckduckgo!

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+dog_licenses %>% filter(animal_name == "BRUNO")
+```
+You could read this code aloud as:
+
+> Take the dog_licenses data, **and then**   
+> filter for only rows when `animal_name` is equal to "BRUNO"
+
+The logical expression, `animal_name == "BRUNO"`, will be `TRUE` when the value of the `animal_name` column is exactly equal to the character string `"BRUNO"` - any differences in case, or whitespace will result in `FALSE`.
+
+Columns names can appear on either side of `==` or even on both sides.  When column names appear on both sides, the comparison happens row by row.  In programming terms this is know as an element-wise comparison.  For example, `license_issued_date == animal_birth_month` will return `TRUE` for a row only if for that row the date the license was issued is the exact same date as the birth month for the dog. If you take a look:
```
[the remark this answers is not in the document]
cwickham: Great suggestion!  First one taken as is.  Second one expanded to include some discussion of what can go wrong when you use quotes around a column name, partially addressing later comments in error section.

r-data-manipulation.Rmd
cwickham: Thanks integrated in.

r-data-manipulation.Rmd
```diff
@@ -0,0 +1,1140 @@
...
+
+* The dog licenses that were issued before 2016
+* The dog licenses that expire before 2016
+
+### More complicated expressions
+
+If you pass more than one logical expression to `filter()` it will return only rows where **all** of the expressions are `TRUE`.  For example, licenses that are to dogs named Bruno in Brooklyn:
+```{r}
+dog_licenses %>% 
+  filter(animal_name == "BRUNO", borough == "Brooklyn")
```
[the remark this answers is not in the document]
cwickham: Good point, I think combining operators is more generalizable too.  I'll rework to use `&`

r-data-manipulation.Rmd
cwickham: Joins removed for now. Conversation moved to #186

r-data-manipulation.Rmd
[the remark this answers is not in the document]
cwickham: This will require some re-writing if we move it earlier to use only functions learners have seen.  Leaving as-is for this PR, and creating an issue to re-visit. See  #188

cwickham:
@lwjohnst86 @gvwilson I think this is ready to go, would one of you please merge?

Issue #142, opened 2019-07-09
Intermediate section: Publishing 


DamienIrving:
This issue tracks the publishing section in the intermediate material.

This issue can be closed when

- [ ] There are an initial set of exercises corresponding to the outline of the material (which may evolve as part of this task)
- [ ] There is a first draft of the text of the section

DamienIrving:
The relevant draft pull request is #143.

PR #143, opened 2019-07-09
revisions to publishing

DamienIrving:
See #142 for details.


cwickham:
I also took a read through and it looks good to me!  

At some point we probably want to link back to the novice materials, e.g. I think by the end of the novice material some could produce a "reproducible" data analysis script that is hosted on github, but we should point out what is new here, or what is missing in the novice approach. I'm thinking at the novice level they haven't:
* seen make,
* documented the software tools, versions etc. they used, nor
* created any persistent pointer to a specific moment in the project history.

rse-publish.Rmd
```diff
@@ -0,0 +1,544 @@
...
+modern publishing involves much more than simply producing a written report.
+It involves providing readers with the data underpinning the report,
+as well as any code written in analysing the data.
+
+While the definition of data (and its associated metadata)
+is relatively easy to understand,
+code can come in many different forms.
+Here we distinguish between “analysis code” written
+solely for the purpose of the report (e.g. to produce a figure)
+and “analysis software” that is formally packaged and released for use by a wider audience.
```
gvwilson: Mm - maybe "script" rather than "code"? There's gotta be better terminology somewhere ...

rse-publish.Rmd
```diff
@@ -0,0 +1,544 @@
...
+
+An [ORCID](glossary.html#orcid) is an Open Researcher and Contributor ID.
+You can [get an ORCID][orcid] for free,
+and you should include it in publications
+because people's names and affiliations change over time.
+
+
+## Publishing a written document {#publish-written}
+
+The best option for publishing a written report (with a platform that issues a DOI)
```
gvwilson: why "written"?

rse-publish.Rmd
```diff
@@ -0,0 +1,544 @@
...
+depends on how your workflow is constructed. 
+If all of a program's parameters are in a configuration file (Chapter \@ref(configure)),
+then that file can be archived.
+Alternatively, you might need to have your program print out its configuration parameters
+and then use `grep` or a script to extract information from the logfile (Chapter \@ref(logging)).
+
+If your workflow involves executing a series of command line programs,
+then you can keep a log/record of the command line entries
+required to produce a given result.
+For example, the [cmdline-provenance](https://cmdline-provenance.readthedocs.io/en/latest/)
```
gvwilson: How widely used is this?

rse-publish.Rmd
DamienIrving: I've only ever seen use of the terms software and code (e.g. [example](https://www.nature.com/articles/d41586-018-02741-4)), but it's not like there's an accepted, widely used language. I like "analysis scripts" better than "analysis code", so I'll swtich to that for now.

rse-publish.Rmd
DamienIrving: Not widely at all. It's a little library I wrote for documenting my own data processing steps. I'd love to include some more widely used examples here - I just don't know what those are.

rse-publish.Rmd
DamienIrving: Good point. I was trying to distinguish the document with the words in it from all the supplementary stuff (code, software, dataset), but saying "written document" sounds a little silly. I've now removed that language.

PR #144, opened 2019-07-10
Start of data wrangling for CO2 and dog data

### LINKED ISSUE #135: Convert column names in datasets to camel case
As per meeting. So they immediately work with clean data.

Later we could get the learners to tidy up a messier more realistic dataset.

lwjohnst86:
@cwickham here's the start of the wrangling. Have at it :)

Related and closes #135


lwjohnst86:
Datasets now available :)

PR #170, opened 2019-08-06
Py intro suggestions

joelostblom:
@ChristinaLK Sorry for not commenting on this in time when you had your PR open initially. I just read through it and here are some suggestions.


py-getting-started.Rmd
```diff
@@ -60,29 +60,30 @@ Python has many built-in functions, that allow you to do different things:
...
+Can you imagine a situation where continuing to type code into the console 
 could become tedious or challenging? 
 
 Here are some examples: 
 
 * Wanting to run the same piece of code many times (you 
 *can* use the up-arrow to see previous commands - try it! - but if you 
 run many commands, your history will quickly become more cluttered. 
-* Needing to re-run a piece of code
-* Most important - when you close Spyder, everything you've typed into the
```
joelostblom: This should not be the case. IPython history persists between sessions by default. I checked with Spyder on my machine and it seems to be working as it should.

py-getting-started.Rmd
```diff
@@ -106,10 +107,35 @@ it becomes quick and easy to run code that's written in a script. Try "running"
 the lines in your script. You should see their output in the console on the 
 bottom right. 
 
+### Organizing Sections with Code Cells
```
joelostblom: I think it is important to introduce this early, since it is integral to how to think about organization within scripts (imo) and analogous to R code chunks.

joelostblom:
I also just noticed that Spyder4 will be released soon and they [renamed a few of the panels](https://github.com/spyder-ide/spyder/wiki/Beta-version-changelog). Nothing that is mentioned in this file but worth keeping in mind going forward.

py-getting-started.Rmd
```diff
@@ -241,33 +267,37 @@ terms that you could search on?
...
+
+    ```{python eval=FALSE}
+    import pandas
+    data = pandas.read_csv("measurements.csv")
+    ```
+
+    Which pane in Spyder will show our new `data` object? What happens when you click on it? 
 
 
 ## Key Points {#rse-r-package-keypoints}
```
lwjohnst86:
```suggestion
## Key Points {#py-getting-started-keypoints}
```

py-getting-started.Rmd
ChristinaLK: Ah, I didn't test this, I just assumed. Good to know.

py-getting-started.Rmd
ChristinaLK: this is great - I only know about cells in the context of a notebook and haven't done it from Spyder.

PR #172, opened 2019-08-08
Add section on python development

### LINKED ISSUE #120: Novice Python section assignment: Development
This issue tracks the [Development](https://github.com/merely-useful/merely-useful.github.io/blob/ab6b9e3a2cbad6c90bc128965bb9b403cfc88201/novice-goals.Rmd#L81-L99) section for the novice Python material.

This issue can be closed when:
- [ ] There are an initial set of Python exercises corresponding to the outline
- [ ] There is a first draft of the text of the section

joelostblom:
I followed the outline, some section might benefit from being moved, merged, or expanded. Mostly thinking about the packages section that maybe should include Python instructions also, and the getting help online section which could be merged with a getting help locally section if we have one of those elsewhere. There is also overlap between the function section and what is in `style.md`.

close #120


py-development.Rmd
```diff
@@ -0,0 +1,520 @@
...
+```{python}
+sum_two_numbers(2, 5, return_input=True)
+```
+
+Since the arguments are given in the same order as in the function definition,
+we could have left out the `return_input=` part and just written `True` in the
+third position. If we want to supply the arguments out of order, we need to
+specify the parameter name of each argument.
+
+#### Function documentation {py-function-documentation}
```
joelostblom: I am not including the Python type annotations. I guess that would be best practice since it allows IDEs to autodetect easier, but I find it quite messy with default arguments (and I believe many IDEs can read numpy doscstring for types).

py-development.Rmd
```diff
@@ -0,0 +1,520 @@
...
+We recommend using the `numpy` docstring format since this is easy to read
+complex function with many arguments and heavily used by the data science
+Python packages. Docstring are what show up in the function help messages, so
+it is important that these are well-written and helpful for the user.
+
+```{python}
+help(sum_two_numbers)
+```
+
+### Applying what we have learnt {py-applying-what-we-have-learnt}
```
joelostblom: It would be great if they have written some small script previously in the course that we could modularize into functions here.

py-development.Rmd
```diff
@@ -0,0 +1,520 @@
...
+```{python}
+sum_of_numbers, len_of_numbers = sum_and_len_all_numbers([1, 2, 3])
+sum_of_numbers
+```
+
+```{python}
+len_of_numbers
+```
+
+> Note: When we defined the function `sum_two_numbers()`, we referred to `num1`
```
joelostblom: This is a suggestion for how to include more advanced tips. We should probably make a nicer div (maybe bootstrap like green or blue color), but looking to get feedback on the idea before settling on design.

py-development.Rmd
```diff
@@ -0,0 +1,520 @@
...
+```
+
+> Note: Similar functionality can be achieved by explicitly raising an error
+> within a `try` and `except` block, or within a conditional statement using
+> `if`, `elif`, and `else`, but `assert` is a simple and readable way of
+> allowing for manual error checking.
+
+## Packages {py-packages}
+
+### Installing Python {py-installing-Python}
```
joelostblom: The following two sections are related to packages and especially package managers so I included them here for now.

joelostblom:
Tagging all (I think) Python novice people @ChristinaLK @mbonsma @gvwilson @dhavide @brandeism @SaraMati @elliewix . Does anyone have time (and wants) to review this?

gvwilson:
Hi @SaraMati - can you please take a run through this one?

py-development.Rmd
lwjohnst86: Yea, not sure how we can do that across output document types (pdf vs ebook vs html). If you have a solid idea, let me know!

Also, I like this parahraph :+1:

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+    '''
+    Add two numbers
+
+    Args:
+        num1 (int, float): The first number to be added.
+        num2 (int, float): The second number to be added.
+
+    Returns:
+        int, float: Sum of the two numbers.
+    '''
```
lwjohnst86: Is there an easy way to add this docstring? E.g. like a key binding (for instance, Ctrl-Shift-R or whatever in RStudio for using roxygen comments).

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+    num1: int, float
+        The first number to be added.
+    num2: int, float
+        The second number to be added.
+
+    Returns
+    -------
+    int, float
+        Sum of the two numbers.
+    '''
```
lwjohnst86: Does this render into a nice format or this is how it looks in the help? Does indentation matter here? Does parameter order matter?

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+help(sum_two_numbers)
+```
+
+### Applying what we have learnt {#py-dev-applying-what-we-have-learnt}
+
+Now we can apply what we have learnt about functions on the code we have
+written previously.
+
+## How to make programs indicate that something has gone wrong?
+{#py-dev-how-to-make-programs-indicate-that-something-has-gone-wrong?}
```
lwjohnst86: This should be on the previous line

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+
+When performing programmatic data analysis, we need to both ensure that our
+code runs correctly and that it carries out the intended tasks. The Python
+interpreter will help us with the first part; if a part of the code is not
+valid, an error will be raised with a message that helps us trace back what
+part of the code is not correct.
+
+### Ducktyping - relying on Python to detect unexpected behavior {#py-dev-ducktyping
+--relying-on-Python-to-detect-unexpected-behavior}
```
lwjohnst86: This should be on the previous line.

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+
+```{python error=TRUE}
+sum_two_numbers(5, 'six')
+```
+
+The raised error contains a helpful message that alerts us to what went wrong;
+`string` objects cannot be added to `int` objects in Python. If the input
+variables behave correctly then they probably are of the correct types and no
+explicit checking is needed. This approach is often referred to as "ducktyping"
+because makes assumption on the type of variable based on its behavior, just
```
lwjohnst86:
```suggestion
because it makes assumptions on the type of variable based on its behavior, just
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+like the saying
+
+> If it looks like a duck, swims like a duck, and quacks like a duck, then it
+> probably is a duck.
+
+### Assertions - explicitly checking for unexpected behavior {#py-dev-assertions--
+explicitly-checking-for-unexpected-behavior}
```
lwjohnst86: previous line

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+
+Ducktyping is useful to catch anything that raises an error in Python, but
+sometimes we might want to stop the code execution for other reason than a
```
lwjohnst86:
```suggestion
sometimes we might want to stop the code execution for reasons other than a
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+technical Python error. Examples of this includes when you have performed a
```
lwjohnst86:
```suggestion
technical Python error. Examples of this include when you have performed a
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+line of code is executed (if there is one).
+
+```{python}
+assert 0 == 0
+print('This is only printed if the assertion above is "True"')
+```
+
+As mentioned above, this is useful if we know that a variable should look a
+certain way, since we can assert if this is the case and guard ourselves from
+errors that origin early in the pipeline but could give rise to more cryptic
```
lwjohnst86:
```suggestion
errors that originate early in the pipeline but could give rise to more cryptic
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+> within a `try` and `except` block, or within a conditional statement using
+> `if`, `elif`, and `else`, but `assert` is a simple and readable way of
+> allowing for manual error checking.
+
+## Packages {#py-dev-packages}
+
+### Installing Python {#py-dev-installing-Python}
+
+There are several ways of installing Python on your system. One of the most
+robust and cross-platform compatible is to install the Anaconda Python
```
lwjohnst86: robust what?

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+
+> Note: You will sometimes hear packages refers to as "modules". The two
+> words are often used interchangeably. Technically, a package is a folder that
+> contains modules (scripts), which in turn contains functions (code).
+
+People around the world have created packages for Python and made them freely
+available for others to use resulting in one of the richest package ecosystems
+for any programming language, with packages for web design, prose writing, game
+development, and data science (to name just a few). Since there are so many
+packages available, it is infeasible to include all of them with the default
```
lwjohnst86:
```suggestion
packages available, it is not feasible to include all of them with the default
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+mn([1, 2, 3])
+```
+
+Which of these you use is up to you, but it is common to follow the conventions
+established by the library's authors, which for numpy is `import numpy as np`
+and use functions via `np.<function_name>`. One thing to avoid is to import
+everything from a package, e.g. `from numpy import *`. If this is done with
+every package, it is almost guaranteed that the same function name will be
+available from more than one package and it will be difficult to keep track of
+if you are using the `mean` function from `numpy` or from another package that
```
lwjohnst86:
```suggestion
it if you are using the `mean` function from `numpy` or from another package that
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+   search -t conda <package name>`, to find a channel with the desired package.
+   To install this package, use `conda install -c <channel name> <package
+   name>`. The [conda forge channel](https://conda-forge.github.io/) channel
+   has many of the packages not in the default repositories.
+
+## How to get help online {#py-dev-how-to-get-help-online}
+
+When reading the built-in Python help is not sufficient, there are several 
+online resources that can helpful. One of the most commonly used resources for
+data science is the Stack Exchange network wich offer Q&A sites both for
```
lwjohnst86:
```suggestion
data science is the Stack Exchange network which offer Q&A sites both for
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+programming related topics via [Stack Overflow][so], as well as statistics and
+machine learning via [Cross Validated][cv].
+
+If nothing relevant can be found after searching the many existing question and
+answers on these sites, it is appropriate to ask a new question! Stack Overflow
+has detailed instruction on how to create a [minimal reproducible
+example][so-mre] when asking a question to increase the chances that the
```
lwjohnst86: In R there is the repro package to automatically create a MRE, rendered into the markdown form needed for SO. Is there something similar in Python?

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+question receives a specific and helpful answer. The key principles listed on
+the website recommends that and answer follows these guidelines:
```
lwjohnst86:
```suggestion
the website recommends that an answer follows these guidelines: 
```

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+referred to as [rubber duck debugging][wiki-rubber-duck-debugging].
+
+## Key Points {#py-dev-keypoints}
+
+```{r child="keypoints/py-development.md"}
+```
+
+## Exercises {#py-dev-exercises}
+
+1. Write a function that returns the mean of exactly two numbers.
```
lwjohnst86: It might help to create some code scaffolding for this exercise.

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+exactly which region of the code is failing and the act of explaining the
+problem often reveals a solution before any replies have come in (commonly
+referred to as [rubber duck debugging][wiki-rubber-duck-debugging].
+
+## Key Points {#py-dev-keypoints}
+
+```{r child="keypoints/py-development.md"}
+```
+
+## Exercises {#py-dev-exercises}
```
lwjohnst86: It might help if these exercises either had more code scaffolding or were written in a way to achieve a specific output based on an input. As they are now, they are too vague.

py-development.Rmd
```diff
@@ -0,0 +1,540 @@
...
+
+1. Write a function that returns the mean of exactly two numbers.
+2. Write a function that returns the mean of any amount of numbers.
+3. Write function that takes a string as its input and returns the
+   first and last character of the string.
+4. Write a function that returns any character from the string. Which character
+   is returned should be determined by an integer specified via a named
+   parameter that has the default value of `0`.
+5. Expand on the previous function by include a assert statement that checks if
```
lwjohnst86: If this is a multi-part exercise, make it one exercise with two parts, not two exercises that are linked.

py-development.Rmd
joelostblom: Yes, good point I will add this. There is actually a keybinding, a menu item, and an autocomplete template ([gif here](https://github.com/spyder-ide/spyder/pull/8700))!

py-development.Rmd
joelostblom: I had a brief look, but could not find anything.

objectives/py-development.md
```diff
@@ -0,0 +1,8 @@
+- Describe a few advantages of functions.
+- Create a function that takes both arguments and keyword arguments.
```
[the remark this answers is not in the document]
joelostblom: Hmmm no, but is it clearer to remove "both"? "Create a function that takes arguments and keyword arguments." or maybe "Create a function that takes arguments as well as keyword arguments."

py-development.Rmd
joelostblom: Good point to bring up. Spyder has rich rendering of docstrings treating it similarly (or maybe exactly as) markdown. This means the NumPy-like docstrings work slightly better as the arguments are line separated without adding extra trailing spaces or newlines in the source:

Numpy docstring format:
image

Google dostring format:
image

I also prefer numpy docstring and they are the most common among scientific packages, so I think I will make a harder recommendation for Numpy format, exclude the Google example and replace it with a link, and add a screenshot of the rendered docstring.

In PyCharm, I believe both renders as rich text, in jupyterlab the raw string is shown.

py-development.Rmd
joelostblom: [This seems promising to me](https://stackoverflow.com/a/47294235/2166823), what do you think? We could create a function for "Advanced notes/Deep dives" (and maybe also "Exercises" if we want that to be more than a standard heading) that authors could call when inserting a similar paragraph as mine here.

Although this now appears technically feasible, and I like having a visually distinct element for advanced notes, it appears I am the only one who has written in this style so far (the rest follow your previous suggestion of having everything in paragraphs. So I am inclined to wander down the path of least resistance here and change my writing instead of encouraging everyone else to alter theirs.

py-development.Rmd
joelostblom: I believe the docstrings are parsed the same as markdown, so yes indentation will matter since it makes the parameter name and description render as a markdown definition list with the parameter name in bold typeface. The order doesn't matter for anything that I am aware of but it is convention to keep the order the same as in the function signature.

py-development.Rmd
joelostblom: buzzZZzword for SEO and to generate hYPeZ among readers.

I can replace with "reliable", "straightforward" or "hassle free", something along those lines.

py-development.Rmd
joelostblom: I think these would actually work, but I changed them so it looks nicer.

objectives/py-development.md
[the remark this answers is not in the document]
lwjohnst86: Well, no, but a keyword argument is an argument too. So why mention both? Or do you mean positional arguments and keyword arguments? Or do you mean something else?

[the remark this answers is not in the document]
joelostblom: Yes sorry, I am referring to positional arguments. The abbreviations in Python are "args" and "kwargs" respectively so I forgot to clarify by prefixing with "positional", good catch!

py-development.Rmd
joelostblom: The scaffolding for this is in the chapter where we walk through how to create a function that calculates the sum of two numbers and the sum and length of an arbitrary long list of numbers. I can add a reminder to revisit that section.

py-development.Rmd
joelostblom: I have added explicit instruction to start from the functions I had in mind from the examples in the section text.

joelostblom:
Thank you for the comments @lwjohnst86 and @ChristinaLK!

Luke, I have updated based on your feedback, let me know what you think.

Christina, thanks for offering to make a suggestion for the intro, please go ahead and I can integrate and expand it if needed.

py-development.Rmd
lwjohnst86: We can revisit this later. That link looks promising..

lwjohnst86:
Looks great to me!

gvwilson:
Thanks all - I'm happy to merge and resolve conflicts - @joelostblom are we good go to?

joelostblom:
@gvwilson Yes let's merge it. @ChristinaLK and I can add a short intro in a separate PR.

gvwilson:
We're missing the file `figures/py-development/docstring-spyder.png` - @joelostblom can you please add to the PR?

joelostblom:
@gvwilson Sorry, added.

Issue #210, opened 2019-10-03
Proposal: move discussion of style to novice material


gvwilson:
## Proposal

Move `style.Rmd` to the novice material.

## Background

`style.Rmd` is currently in the RSE material, but we can and should teach good coding style earlier.

## Pros and Cons

**Pro:**
-   We should teach good programming style as early as possible.
-   Nothing in the material requires in-depth understanding of Python or R.
-   Load balancing (evens out sizes of volumes).

**Con:**
-   Will need to do for both Python ~~and R~~ (*Note*: Already done in r-reproducibility).

## Alternatives

NA.

lwjohnst86:
I think this is a good idea, though I've already done a big section on style in the r-reproducibility section. So it would have to be done for the Python. I don't know how its like in Python, but the `styler` package makes it super easy to adhere to the style guide in an automated way.

lwjohnst86:
The bit about code review would be very useful in the RSE, but not really the novice material though

lwjohnst86:
Passes. Move into novice.

PR #215, opened 2019-10-05
Style suggestions

joelostblom:
@gvwilson Since there are a few bigger changes I felt more comfortable with a PR rather than directly merging.


joelostblom:
@gvwilson Also, on line 442, what is a "point object"?

lwjohnst86:
Looks good. Given the #210 proposal, I'll merge and edits can be done later (or moved over into the Novice Python section).

Issue #156, opened 2019-07-23
Revise section on configuration of software


ljdursi:
Revise `configure.Rmd`:

- [ ] Exercises
- [ ] Is an R section needed?

DamienIrving:
See #223

DamienIrving:
As it stands, `configure.Rmd` covers how to write a python script (using argparse) and then looks at (and implements) overlay configuration. 

I was initially assigned to this issue and started working on an update to `configure.Rmd` but wasn't sure how relevant the overlay configuration content is for our audience. At the [meeting on 8 October](https://github.com/merely-useful/merely-useful.github.io/wiki/Meeting-agenda-2019-10-08) we decided to move `configure.Rmd` to `configure.md` for now and to discuss further how best to teach configuration (if at all). The beginnings of that discussion at at #223 but we should continue the discussion here.

(Python scripting will now be a separate chapter.)

gvwilson:
@DamienIrving closing snce we have remove configure.Rmd

PR #223, opened 2019-10-07
Revisions to RSE configuration chapter

DamienIrving:
See #156.


DamienIrving:
In this chapter we cover how to write a python script (using argparse) and then look at (and implement) overlay configuration. I have lots of experience with python scripts but not with all the different types of configuration files. 

Do others use configuration files extensively in their work? (In other words, is it weird that I don't use them, or are they not something that our intermediate learners need to know about?)

DamienIrving:
Also, when I use `make html` the formatting for this lesson is all messed up and I have no idea why.

lwjohnst86:
@DamienIrving I don't use configurations and, at least for R packages and programming, configuration doesn't make sense/is rarely used. So I would vote to not have this in it, unless the Python users are big users of it.

joelostblom:
I don't use configurations files much either. I would be inclined to say
it's enough to show how to setup variables on top of a script and mention
that these could be broken out into their own python file or a general text
format such as yaml or json which could then be parsed (and include a link
to read more).

I think it would be nice if we somewhere included heuristics for when to go
from scripts to writing functions and packages. Not necessarily here but
thought of it because of the similarity to deciding when to break out
variables to config files.

cwickham:
It occurred to me during the call today, that two common ways R packages expect users to be able to set configuration parameters is through editing YAML files (e.g. bookdown, Rmarkdown) or setting environment variables (e.g. gh, ?).  How exactly they are used is probably so package specific that it isn't of wide interest, although some exposure, particularly to yaml and why people use it, might be beneficial.

Another way to frame the configuration stuff would be: how do I set up project wide (or user wide) settings/variables?

lwjohnst86:
100% agree with @cwickham, when I hear "configuration", I think of things like configuration files for Bash programs or other programs. But "project level settings" is much more intuitive from a data analysis perspective, at least to me.

lwjohnst86:
@DamienIrving did you figure out the `make html` problem you were experiencing? I just tried it out and it worked fine.

DamienIrving:
Thanks, @cwickham and @lwjohnst86. That's super helpful. At this stage I think I'll close this PR and we should continue the discussion of what (if anything) should be taught regarding configuration at #156.

I'll open separate PRs soon to (a) move `configure.Rmd` to `configure.md` and (b) for a stand-alone lesson on python scripts.

Issue #220, opened 2019-10-06
rse-bash suggestions


joelostblom:
@gwilson Aside from my suggested changes in the PR #219 , here are a few comments for you to do with as you see fit.

- [ ] Remove `$` code prefix everywhere
    - This will make code easier to copy
    - We don't include the prompt character for either Python or R.
- [ ] Where do we instruct readers to download the folders and follow the setup
      instructions for the carpentry bash lecture?
- [ ] Should we mention `.basrhc` instead of `.bash_profile` as an example of
  a hidden file since we mention that this is the config file, but I persume
  that we will modify `bashrc` if we are to show how to configure bash?
  I didn't make the changes because I wasn't sure if macOS for some reason did
  not always have `.bashrc`,but `bash_profile` instead. - 
- [ ] I am used to writing `n`/`N` instead of `N`/`Shift+N`, but maybe that's
  a vim thing.

Things that I did not see mention of that could be useful
- &, ^Z, and fg.
- sudo
- shebang line
- aliases
- bashrc config
- the escape character (\\)
- {a,b} for matching either or
- single vs double quotes

ChristinaLK:
+1 to single vs double quotes -- it seems picky, but when you use variables it really does matter. 

+1 also to shebang line -- generally useful and also good to clarify that it's not a comment.

gvwilson:
Thanks for the feedback:

- Remove `$` for consistency: yup.
- Download instructions: web page rather than in the book, because instructions will change more frequently than content (particularly during first contact with users).
- Single quotes: check.
- shebang: only useful if we teach permissions, which we don't.
- sudo, aliases, etc.: sure, as long as you can tell me what we take out to make room, because this is already the longest chapter in the whole series by a factor of 2.

joelostblom:
> +1 to single vs double quotes -- it seems picky, but when you use variables it really does matter.

Yes, especially since it does not matter in R and Python, I think it is important to just mention this in a short sentence, possibly with a link to a more complete explanation of variable expansion. 

> sudo, aliases, etc.: sure, as long as you can tell me what we take out to make room, because this is already the longest chapter in the whole series by a factor of 2.

Point well taken. The reason I bring up aliases and bashrc is that I want to mitigate the annoyance that can be felt from having to type the same flag every time a command is used (We also mention "the shell's default settings" in the section without explaining what that means). I think this is important and powerful enough to justify a few sentence along the lines of:

[...]

PR #229, opened 2019-10-09
Major overhaul of RSE Bash chapter

gvwilson:
1. Major rewrite of `rse-bash.Rmd` - probably easiest to read as-is rather than viewing diff.
2. Minor edits to other chapters along the way.
3. Moved exercise solutions into `rse-solutions.Rmd`.
4. Added `climate-data` directory to use in Bash examples - includes some Australian climate data.
5. Replacing direct links to glossary entries with double square bracket links in `links.md`: makes main text more readable.


rse-bash.Rmd
```diff
@@ -913,339 +634,538 @@ draft.txt
...
+> To search for a character or word,
+> use <kbd>/</kbd> followed by the character or word to search for.
+> If the search produces multiple hits,
+> we can move between them using <kbd>N</kbd> (for "next").
+> To quit, press <kbd>Q</kbd>.
+
+This is a lot of information,
+most of which isn't really useful.
+Some commands have a `--help` option to provide a succinct summary of possibilites,
+but the best place to go for help these days is probably the [TLDR][tldr] website.
```
joelostblom: There is also a tldr shell program one can use instead/in addition to of man pages.

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
+
+```text
+cat: all.txt: No such file or directory
+```
+
+Every process has a second output channel called [standard error][standard-error] (or [stderr][stderr]).
+Programs use it for error messages
+so that their attempts to tell us something has gone wrong don't vanish silently into an output file.
+There are ways to redirect standard error,
+but doing so is almost always a bad idea.
```
joelostblom: Isn't it quite useful for logging output from a command (e.g. with tee or straight to file)? Bash somewhat recently introduced `|&` as an abbreviation for `2>&1 |`, so I would believe it sees a fair amount of use to justify a shortcut.

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
+and a simple way to repeat it
+are yet another reason why people use the Unix shell.
+In fact,
+being able to repeat history is such a powerful idea
+that the shell gives us several ways to do it.
 
-We would like to modify each of the files in `data-shell/creatures`, but also save a version
-of the original files, naming the copies `original-basilisk.dat` and `original-unicorn.dat`.
-We can't use:
+-   `!head` re-runs the most recent command starting with `head`,
```
joelostblom: This is where zsh is so nice because it will exapnd `!head` and show the full command =)

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
-# Usage: bash middle.sh filename end_line num_lines
-head -n $2 "$1" | tail -n $3
+# Select distinct years from column 2 of climate data file.
+# Usage: bash years.sh /path/to/file.csv
+cut -d , -f 2 $1 | uniq
 ```
 
-A comment starts with a `#` character and runs to the end of the line.
+As in R and Python,
+a comment starts with a `#` character and runs to the end of the line.
```
joelostblom: Nice that we picked three languages with the same comment character  =)

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
 ```text
-# Sort filenames by their length.
-# Usage: bash sorted.sh one_or_more_filenames
-wc -l "$@" | sort -n
+Station
+68228
 ```
 
+But we have made a common mistake:
+we have changed the script without changing the comment.
```
joelostblom: Great to bring this up!

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
+A description that sends readers in the wrong direction is worse than none at all,
+so we should go back and update it.
+We should probably also change the script's name from `years.sh` to `column.sh`,
+since a program's name is the first piece of "documentation" anyone sees.
+
+We should probably also add one more command to our pipeline.
```
joelostblom: Same wording as previous sentence

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
-297 bash goostats NENE01729B.txt stats-NENE01729B.txt
-298 bash goodiff stats-NENE01729B.txt /data/validated/01729.txt > 01729-differences.txt
-299 cut -d ',' -f 2-3 01729-differences.txt > 01729-time-series.txt
-300 ygraph --format scatter --color bw --borders none 01729-time-series.txt figure-3.png
-301 history | tail -n 5 > redo-figure-3.sh
+297 bash stats.sh cleaned/*_temp.csv > temperature_stats.csv
+298 bash trim-outliers.sh temperature_stats.csv > plot_data.txt
+299 date
+300 ygraph --format scatter --color bw --borders none plot_data.txt figure-3.png
+301 rm temperature_stats.csv plot_data.txt
```
joelostblom: Maybe make these the last 6 commands in this lesson? so `cut` etc

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
-what they discover about their data and their workflow with one call to `history`
-and a bit of editing to clean up the output
-and save it as a shell script.
-
-## Finding Things {#rse-bash-finding}
+It only takes a few moments in an editor to remove the serial numbers
+and delete the use of `date`
+(which prints the current time and date)
+to create a script that accurately captures what we actually did.
+This is how we usually develop shell scripts:
```
joelostblom: Tricks like this about practical use are great and I think we shuold have them more often in the other lectures also

rse-bash.Rmd
```diff
@@ -2160,342 +2209,284 @@ $ grep -n "it" haiku.txt
...
 9:Yesterday it worked
-10:Today it is not working
 11:Software is like that.
 ```
 
-To search for text within multiple files in the same directory,
-wildcards can be used when specifying the file names.
-For example,
-`grep "the" *.txt` would search all `.txt` files in the current directory.
-To search all files within a directory and its subdirectories,
```
joelostblom: I would add these 4 lines back in. `grep -r` is useful and often needed in my experience

rse-bash.Rmd
```diff
@@ -1256,801 +1176,895 @@ $ ls lengths.txt
...
 ```shell
-cp unicorn.dat original-unicorn.dat
+$ cp cleaned/*_prec.csv backup/*_prec.bak
 ```
 
-Before running a loop,
-it can be a good idea to check that the loop is doing the correct thing.
-We learned earlier how to print strings using `echo`,
-and we can modify the loop
-to use `echo` to print our commands without actually executing them.
```
joelostblom: I think this is a great practical tip to add back in

joelostblom:
Oh and I think it is important to briefly point out that we must use `"` here altough R and Python allows `'` to have the same meaning.

OUTTAKES.md
```diff
@@ -30,3 +32,28 @@ we will see if the similarity measures we develop show that.
 | Bram Stoker                 | dracula.txt                     |
 | H. G. Wells                 | time_machine.txt                |
 | Edith Wharton               | ethan_frome.txt                 |
+
+---
+
+# Editor Recommendations
+
+However, because of this trait, it may
```
lwjohnst86: What does "this trait" refer to? And is this in reference to Python or are you including RStudio? At present, I don't think there is anything as powerful as RStudio for R, but I don't see that mentioned here.

climate-data/climate-data/bin/slice.R
```diff
@@ -0,0 +1,17 @@
+library(tidyverse)
+
+args <- commandArgs(trailingOnly=TRUE)
+firstYear <- as.numeric(args[1])
+lastYear <- as.numeric(args[2])
+filenames <- args[c(-1, -2)]
+
+slice <- function(name, from, to) {
```
lwjohnst86: There's already a function called slice in dplyr. Does that not do what you want?

lwjohnst86: Not that this part is relevant to change, since I assume this is for getting data ready for the lesson.

rse-bash.Rmd
```diff
@@ -1,906 +1,627 @@
...
+and interact with people.
+They can do the last of these in many different ways,
+of which a [graphical user interface][gui] (GUI) is the most widely used way.
+The computer displays icons to show us our files and programs,
+and we tell it to copy or run those by clicking with a mouse.
+This style of interaction is easy to learn but hard to automate,
+and doesn't create a record of what we did.
+
+In contrast,
+when we use a [command-line user interface][clui] (CLUI)
```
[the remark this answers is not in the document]
lwjohnst86: Definitely agree here, google CLUI and first few show up as land use. Google CLI and first entry is relevant.

rse-bash.Rmd
```diff
@@ -1,906 +1,627 @@
...
-*do not type the prompt*, only the commands that follow it.
+However,
+different shells may use a different symbol,
+and as we'll see later,
+we can customize the prompt to give us more information.
 
-So let's try our first command, which will tell us our username:
+> **Don't Type the Dollar Sign**
+>
+> We show the `$` prompt so that it's clear what you are supposed to type,
```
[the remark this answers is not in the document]
lwjohnst86: I 100%... no a million % agree with this. We don't do this with R or Python, why do it with Bash? Plus it is a major pain in the ass to copy and paste from.... As we agreed in #220, let's remove it.

rse-bash.Rmd
```diff
@@ -913,339 +634,538 @@ draft.txt
...
+`wc` has other options as well.
+We can use the `man` command (short for <strong>man</strong>ual)
+to find out what they are:
+
+```shell
+$ man wc
+```
+
+```text
+WC(1)                     BSD General Commands Manual                    WC(1)
```
lwjohnst86: Is this full text necessary for the lesson? Could it be shorter?

rse-bash.Rmd
```diff
@@ -1,906 +1,627 @@
...
+and the computer responds by displaying text.
+CLUIs existed long before GUIs,
+and have survived because they are efficient,
+easy to automate,
+and automatically create a record of what we did.
+
+The heart of every CLUI is a [read-evaluate-print loop][repl].
+Its name comes from its basic operating cycle:
+when we type a command and press <kbd>Return</kbd> (also known as <kbd>Enter</kbd>)
+the CLUI <strong>r</strong>eads the command,
```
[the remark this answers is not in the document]
gvwilson: Added `strong` tags around the other three letters - I use `strong` instead of `**` because the latter isn't parsed reliably.

rse-bash.Rmd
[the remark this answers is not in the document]
gvwilson: 1. I think we need to show prompts when we introduce multi-line commands (`for` loops).
2. I think we also need to show them when we have two or more users in play (which happens in the Git chapters, where the prompt includes the user ID).
3. I think we _should_ show the prompt for Python/R when we are asking people to work directly in the console rather than writing a script, so that they can more easily track how we're doing things - in this chapter, we _don't_ show the prompt when they are writing a script.
4. Looking at books by O'Reilly, and the original Kernighan et al books, they show the prompt for interactive shell sessions.

I have filed a separate discussion issue and we can re-discuss at the next meeting - please see #236.

rse-bash.Rmd
gvwilson: thanks - fixed

rse-bash.Rmd
gvwilson: See the exercise "Doing a dry run".

OUTTAKES.md
gvwilson: I moved this to `OUTTAKES.md` because it needs to go somewhere, but I'm not sure where - a chapter or appendix on "other tools"?

climate-data/climate-data/bin/slice.R
gvwilson: Yes, but I fixed it anyway - good catch.

lwjohnst86:
I'll merge it in. I'll probably take a closer look at it when other chapters been put together.

Just FYI, you can use knitr to generate the output of the shell commands, rather than you copy and pasting from the shell. Adding the knitr engine "bash" lets you do that. (if you did that you'd have to remove the `$` prompt though :wink:)

PR #232, opened 2019-10-10
Intermediate chapter on Python scripting

DamienIrving:
Note: The content on Python scripting will be separate from anything on configuration (see #156).


rse-py-scripting.Rmd
```diff
@@ -0,0 +1,143 @@
...
+
+## Introduction {#rse-py-scripting-intro}
+
+The Jupyter Notebook, PyCharm and other interactive tools
+are great for prototyping code and exploring data,
+but in many cases you will ultimately want to take the code you've developed
+using those tools and apply it to thousands of data files, 
+run it with many different parameters,
+and/or combine it with other programs in a data analysis pipeline. 
+In order to do these things,
```
joelostblom:
```suggestion
To effectively perform these tasks,
```

rse-py-scripting.Rmd
```diff
@@ -0,0 +1,143 @@
...
+it is often necessary/preferable to turn the code you've developed into a program
```
joelostblom:
```suggestion
it is often necessary to turn the code you've developed into a program 
```

rse-py-scripting.Rmd
```diff
@@ -0,0 +1,143 @@
...
+
+The Jupyter Notebook, PyCharm and other interactive tools
+are great for prototyping code and exploring data,
+but in many cases you will ultimately want to take the code you've developed
+using those tools and apply it to thousands of data files, 
+run it with many different parameters,
+and/or combine it with other programs in a data analysis pipeline. 
+In order to do these things,
+it is often necessary/preferable to turn the code you've developed into a program 
+that works like any other Unix command-line tool.
```
joelostblom:
```suggestion
that can be run from the Unix shell as a standalone program just like other command-line tools.
```

rse-py-scripting.Rmd
```diff
@@ -0,0 +1,143 @@
...
+
+In general, the first thing that gets added to any Python script is the following:
+
+```python
+if __name__ == '__main__':
+    main()
+```
+
+The reason we need these two lines of code
+is that running a Python script in bash is very similar to importing that file in Python.
```
joelostblom: I remember struggling with this when I started learning Python... I think we could make these two paragraphs slightly clearer, what do you think of something along these lines?

```suggestion
The reason we need these two lines of code
is to differentiate between running a Python file as a standalone program
and importing it as a module. 
When you import a Python file as a module,
the `__name__` variable is automatically set to the name of the file
(e.g. when importing script.py, `__name__` is `script`).
When running a Python file as a standalone program,
`__name__` is always set to `__main__`.
This means that we can separate the two cases above by checking the value of `__name__`.
When the Python file is running as a standalone program,
you want to execute its main functionality.
You conventionally define what this is inside a function named `main()`,
but you could call this function whatever you prefer.
```

rse-py-scripting.Rmd
```diff
@@ -0,0 +1,143 @@
...
+
+The next thing you'll need is a library to parse the command line for input arguments.
+The most widely used option is 
+[argparse](https://docs.python.org/3/library/argparse.html).
+
+Putting those together,
+here's a basic python script for us to start with:
+
+```shell
+$ cat script_template.py
```
joelostblom: Do we need this code block? Can we say "Create the file `script_template.py`" instead?

rse-py-scripting.Rmd
```diff
@@ -0,0 +1,143 @@
...
+```shell
+$ python script_template.py in.csv
+```
+
+```text
+usage: script_template.py [-h] infile outfile
+script_template.py: error: the following arguments are required: outfile
+```
+
+## Adding our code {#rse-py-scripting-code}
```
joelostblom: Should we cover optional and default arguments before this or it that outside the scope of this book?

DamienIrving: I think in the process of "adding our code" we should introduce optional and default arguments.

DamienIrving:
@joelostblom Thanks for your initial review - I basically incorporated all the changes that you suggested.

At this point this chapter introduces the basics of python scripting and then has a big placeholder:
> TODO: In this section we will take the code developed in the novice book and turn it into a Python script that can be executed at the command line. See [this lesson](https://carpentrieslab.github.io/python-aos-lesson/04-cmdline/index.html) for an example of this approach in action.

In other words, I was anticipating that there would be a common narrative (including common data and code) running through the novice and intermediate books (or at least the intermediate book). This might not actually be true (I've started a discussion at #240 to figure this out), so we could either merge this PR with the placeholder in it or wait until #240 is resolved and replace the placeholder with actual content?

joelostblom:
Changes look good, and I also think it would be nice with a narrative throughout. I believe this needs a rebase before merging, or is there another reason that there are so many unrelated commits in the diff with book/master?

gvwilson:
The mess of diffs against `book` are my fault - with your permission, I'll merge from book into `pyscripts` and send a PR to @DamienIrving so that this work is all on top of the changes I've been pushing.

joelostblom:
:+1: Sounds good to me!

DamienIrving:
Thanks, @gvwilson. It's looking good now.

Should we merge this and sort out the #240 issue later?

PR #272, opened 2019-11-06
More on data

gvwilson:
A few more points from some papers - needs a lot of editing to integrate, and we may wind up with a separate chapter on data as this grows.


rse-project.Rmd
```diff
@@ -692,6 +692,53 @@ Breaking this down:
...
+    every record should have a unique [key][key] within the dataset (Section \@ref(rse-publish-data)).
+
+-   **Have a backup plan.**
+    Every hard drive eventually fails;
+    the hardware needed to read various storage media eventually becomes unavailable,
+    and there is always the risk that bankruptcy, corporate merger, or a court order
+    will shut down the site we've been using to store our data.
+    We should therefore always have a second copy in a different physical and jurisdictional locational.
+    If the primary store is GitHub,
+    copy the data onto a CD-ROM and give it to the departmental secretary for safekeeping.
```
joelostblom: Not sure how many of the readers will have access to a departmental secretary...

```suggestion
    copy the data to a secondary online storage provider,
    or a CD-ROM and give it to the departmental secretary for safekeeping.
```

rse-project.Rmd
```diff
@@ -692,6 +692,53 @@ Breaking this down:
...
+    (Their filing system is probably more systematic than ours.)
+
```
joelostblom:
```suggestion
If the data contains sensitive information,
such as personal health records,
you must ensure that the storage location complies with the data privacy regulations
and that the data is properly anonymized before it is uploaded to an off-site storage location.

```

rse-project.Rmd
```diff
@@ -692,6 +692,53 @@ Breaking this down:
 > Documentation that serves these two groups well
 > will almost certainly serve the needs of strangers as well.
 
+These additional rules are taken from [@Good2014,@Mich2015,@Hart2016,@Zook2017]:
+
+-   **Keep raw data raw.**
```
elliewix: Raw data may also just the data as you received it or downloaded it from a repository.  Retaining an unchanged version is important for reproducibility, and in case you use the wrong read/write mode and blow away your working copy.

rse-project.Rmd
```diff
@@ -692,6 +692,53 @@ Breaking this down:
...
+    However,
+    our analysis methods and tools can change over time,
+    and even after years of work they can contain bugs.
+    We should therefore keep the original data exactly as we received it
+    so that we can re-do previous analyses or try new ones.
+    This is also a good safeguard if anyone ever questions the integrity of the research:
+    it will be hard to defend ourselves against accusations of fraud
+    if we can't show our work.
+
+-   **Give every part of the data a unique identifier.**
```
elliewix: "every part of the data" is extremely unclear here.  This reads like every cell in a dataset should have a DOI, which is safe to presume not the intent.  Can this be revised a bit?  Every datapoint in a dataset should have an "address" or location description that can uniquely identify it. For example, a data point may be within file ACX-02.csv, under column reading_type, and in row 450. 

I'm not sure I like to see discussion of persistent identifiers and row/column ids within the same discussion.  File names, column names, and entity IDs should be unique, meaningful, and their patterns/meaning documented.  Datasets should have DOIs once used for a publication.

rse-project.Rmd
elliewix: If I tried to do this I would get massive pushback because their filing systems are not designed for data nor is my data their responsibility to ensure.  Many academic institutions have some form of institutional repository where things can be uploaded to, some long term data storage solution, or a unit better equipped to handle long term data storage requests than front office admin staff.  Also, some cheaper brands of burnable CD-ROMs only have a lifespan of 1-2 years.  People should first contact their departmental level IT department to ask for policies, then expand their searches to their library, research IT group, or grant compliance staff as needed.

rse-project.Rmd
joelostblom: Yes that wording was unclear to me at first as well. I also read it as it referred to every cell or variable.

rse-project.Rmd
lwjohnst86: Definitely agree, not sure it is the secretaries job here. They are usually quite busy as it is, and I don't think we should add this to their list of duties. Simply saying "store in a secure location" works just as well.

lwjohnst86: And I'm not sure I would say "CD-ROM". Maybe USB key instead? Or if using a CD-ROM, store only the more "final" version of a dataset? Can't see how this would work in actual practice... Combined with the fact that some data needs to be securely stored (health or personal data). I'd remove reference to CD-ROM.

rse-project.Rmd
lwjohnst86: 100% agree. GDPR here in Europe makes working and storing health/personal/sensitive data quite difficult, so its important to include this mention somewhere.

rse-project.Rmd
elliewix: changes look good

rse-project.Rmd
elliewix: The updates look good, but Luke should take a look first.

elliewix:
I only approved on one file but it marked as both.  The updates look good, with one comment on privacy that Luke should review and resolve.  This is now getting into the depths of github that are beyond me, but what I can see looks ok.

Issue #241, opened 2019-10-24
Add heuristics and workflow advice for when to modularize code


joelostblom:
In the py-dev chapter, I have written about [functions](https://github.com/merely-useful/merely-useful.github.io/blob/book/py-development.Rmd#L8) and [packages](https://github.com/merely-useful/merely-useful.github.io/blob/book/py-development.Rmd#L339). I think this chapter can be reorganized and a few lines can be added to explain criteria for when to break out code into functions and briefly also packages and then point to RSE for more details around creating packages. 

@elliewix Thank you for offering to add the material you have already. Do you think it would fit well in this chapter to extend or bridge the section I mentioned above? As I said, we can reorganize as we see fit, I just followed the order in the outline file to make sure I covered all the bullets.

@DamienIrving Then for the packaging section, you could maybe briefly repeat when we modularize into functions and packages and go into details about packages. Just to make sure, [is this the file](https://github.com/merely-useful/merely-useful.github.io/blob/book/rse-package-py.Rmd) where the material would go? [This blog post](https://florianwilhelm.info/2018/11/working_efficiently_with_jupyter_lab/) talks about some of the principles around extracting code from the notebook into modules (up until section 2-3), which is largely what I am thinking about (have an edible packages and use `%autoreload 2`)

---

Somewhat of and aside, do we want to make a difference between a local package and a package for distribution? In my workflow, I rarely have .py files in the same directory as my notebooks, instead I create a local editable package so that the notebook imports are tidier and don't need to change if I decide to move things around. To be clear the diff would be that I can write

`from mypkg.myfile import myfunc`

instead of 

`from myfile import myfunc` (and then have to change this if I move files around or want to create a package).

This is maybe not a major point, but I wanted to mention it so it is clear how I am thinking about this and point out that local packages like these don't require all the things a public package requires (readme, contrib, coc, version etc), it's more like an organized dump of functions from notebooks (often multiple in the same project). The more I write, the more this sounds like something that should go into RSE instead of novice...

elliewix:
I reviewed the chapter and have made additions.  I was thinking more about adding things about creating your own custom function file, but I see that's pretty well covered in another chapter.  I'll be submitting a PR with my proposed additions. Feel free to accept/squash per your vision.

elliewix:
OK I have made this thing, https://github.com/merely-useful/merely-useful.github.io/pull/273, which may or may not have been done right.

PR #273, opened 2019-11-06
tweaking some content around package importing

elliewix:
For #241


py-development.Rmd
```diff
@@ -407,8 +405,31 @@ packages right from your terminal, let's see how we can use them!
...
+
+Import statements are generally found at the beginning of script or notebook files.  
+Note that many of these examples will talk about importing functions, but there are 
+several other types of content that can be imported from a package. 
+
+There are two main ways of importing content:
+
+1. Import the entire package via `import <package_name>`, giving you access to the 
+package content by way of the package's name. E.g. `<package_name>.<function_name>()`.
+2. Import just content that you want, but directly into your current namespace via
```
joelostblom:
```suggestion
2. `from <package_name> import <function_name>`: Import the entire package, but only define variables for the functions you list explicitly.
```

My understanding is that the entire package is still imported, just that the `foo` variable is not set. From a brief look [this SO answer](https://stackoverflow.com/a/21547583/2166823) was the first reference I could find.

py-development.Rmd
```diff
@@ -407,8 +405,31 @@ packages right from your terminal, let's see how we can use them!
...
+Import statements are generally found at the beginning of script or notebook files.  
+Note that many of these examples will talk about importing functions, but there are 
+several other types of content that can be imported from a package. 
+
+There are two main ways of importing content:
+
+1. Import the entire package via `import <package_name>`, giving you access to the 
+package content by way of the package's name. E.g. `<package_name>.<function_name>()`.
+2. Import just content that you want, but directly into your current namespace via 
+`from <package_name> import <function_name>`. This gives you access to that content
```
joelostblom:
```suggestion
This gives you access to the imported content 
```

py-development.Rmd
```diff
@@ -407,8 +405,31 @@ packages right from your terminal, let's see how we can use them!
...
+code can become confusing if you aren't extremely conversant in many of the packages.  
+This is where good documentation comes in.
+
+Import statements are generally found at the beginning of script or notebook files.  
+Note that many of these examples will talk about importing functions, but there are 
+several other types of content that can be imported from a package. 
+
+There are two main ways of importing content:
+
+1. Import the entire package via `import <package_name>`, giving you access to the
```
joelostblom:
```suggestion
1. `import <package_name>`: Import the entire package, giving you access to the 
```

py-development.Rmd
```diff
@@ -437,8 +461,11 @@ import numpy as np
 
 np.mean([1, 2, 3])
 ```
+Using this syntax makes `numpy` not defined within that code.
```
joelostblom:
```suggestion
Using this syntax makes the variable `numpy` not defined in the current namespace. 
```

Note to self that I am not sure if we have introduced namesspaces at this point, but maybe this is something to sort out during an edit round of the entire novice material.

Hmm

py-development.Rmd
```diff
@@ -407,8 +405,31 @@ packages right from your terminal, let's see how we can use them!
...
+
+There are two main ways of importing content:
+
+1. Import the entire package via `import <package_name>`, giving you access to the 
+package content by way of the package's name. E.g. `<package_name>.<function_name>()`.
+2. Import just content that you want, but directly into your current namespace via 
+`from <package_name> import <function_name>`. This gives you access to that content 
+without having to specify the package name. E.g. `function_name()`.
+
+Each of these methods has a way of renaming the package or content name to shorten
```
joelostblom: I think this paragraph is already covered below where we talk about how to use `as`. If you there are some aspects that are missing, could you add them at the respective locations below?

py-development.Rmd
```diff
@@ -428,6 +449,9 @@ all available function and intelligently filter them as you type out more
 letters, try it! When you start getting familiar with typing function names,
 you will notice that this is often faster than looking for functions in menus.
 
+The `import <package_name>` syntax is the most common and generic method of importing
```
joelostblom: Isn't this covered by the two bullets you added above or are you thinking about additional ways of importing?

py-development.Rmd
```diff
@@ -342,16 +342,14 @@ more in detail elsewhere in this book.
...
 
-Certain functionality that is considered essential for the Python programming
-language is available wherever Python is installed. Other highly useful, but
-often more domain specific functionality can be accessed separately in the
-form of Python packages. A package is essentially a few Python scripts in a
-specific directory structure coupled with installation instructions for the
-computer.
+A package is essentially a few Python scripts in a specific directory structure 
+coupled with installation instructions for the computer. Python packages can 
+come in from a few sources.  Note that you will sometimes hear packages refers
```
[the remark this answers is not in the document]
elliewix: I'd like to keep this more generic to highlight that many packages and modules are built in within python, but you can add more as needed. I've fiddled with the suggestion here if that's ok. Just to add a small piece.

py-development.Rmd
elliewix: I was trying more to preview that it was an option for both.  Would "As described in later sections with the `as` keyword, each of these methods has a way of..." make more sense here?

py-development.Rmd
elliewix: No, I think I added that first before the other section and didn't remember to remove it.  It can be deleted (I don't know how in this view).

py-development.Rmd
```diff
@@ -428,6 +449,9 @@ all available function and intelligently filter them as you type out more
 letters, try it! When you start getting familiar with typing function names,
 you will notice that this is often faster than looking for functions in menus.
 
+The `import <package_name>` syntax is the most common and generic method of importing 
+packages, but there are several other ways to do this.
```
joelostblom:
```suggestion
```

py-development.Rmd
```diff
@@ -428,6 +449,9 @@ all available function and intelligently filter them as you type out more
 letters, try it! When you start getting familiar with typing function names,
 you will notice that this is often faster than looking for functions in menus.
 
+The `import <package_name>` syntax is the most common and generic method of importing 
+packages, but there are several other ways to do this. 
+
```
joelostblom:
```suggestion
```

py-development.Rmd
```diff
@@ -342,16 +342,14 @@ more in detail elsewhere in this book.
...
 
-Certain functionality that is considered essential for the Python programming
-language is available wherever Python is installed. Other highly useful, but
-often more domain specific functionality can be accessed separately in the
-form of Python packages. A package is essentially a few Python scripts in a
-specific directory structure coupled with installation instructions for the
-computer.
+A package is essentially a few Python scripts in a specific directory structure 
+coupled with installation instructions for the computer. Python packages can 
+come in from a few sources.  Many that you will use are part of the standard distribution, but packages can be created by anyone and there are thousands of that can be downloaded and installed from online repositories.  Note that you will sometimes hear packages refers
```
joelostblom:
```suggestion
come from various sources.  Many that you will use are part of the standard distribution, but packages can be created by anyone and there are thousands of that can be downloaded and installed from online repositories.  Note that you will sometimes hear packages referred 
```
Just two typos and avoid "few" in two sentences in a row.

py-development.Rmd
```diff
@@ -407,8 +405,31 @@ packages right from your terminal, let's see how we can use them!
...
+There are two main ways of importing content:
+
+1. `import <package_name>`: Import the entire package, giving you access to the 
+package content by way of the package's name. E.g. `<package_name>.<function_name>()`.
+2. `from <package_name> import <function_name>`: Import the entire package, but only define variables for the functions you list explicitly.
+This gives you access to the imported content 
+without having to specify the package name. E.g. `function_name()`.
+
+Each of these methods has a way of renaming the package or content name to shorten 
+it or in case there are multiple with the same name and you need to avoid namespace clashes.
```
joelostblom:
```suggestion
it or in case there are multiple with the same name and you need to avoid namespace clashes. 
This is done using the `as` keyword, which is described in detail later in this chapter.
```

Issue #240, opened 2019-10-23
Discussion: Linkages between RSE chapters 


DamienIrving:
As it stands right now, there is very little relationship between the data, code and/or narrative used from one RSE chapter to the next:  
* **Bash shell**: Uses a `climate-data` directory
* **Git**: Frances and her colleague Jean Jennings write a history of the ENIAC project
* **Python scripting**: Yet to be decided (#232) 
* **Automation**: The distribution of word frequencies in classic English novels is analyzed using two pre-prepared scripts (`countwords.py` and `collate.py`)
* **Packaging**: Develops some "Zipf code" which builds on the word frequency data and narrative (but not code) from the automation chapter

The other RSE chapters (please correct me if I'm wrong) don't really involve a specific dataset or analysis narrative and don't use any code that was developed in a previous chapter or used again in a subsequent chapter.

I'm opening this issue to discuss whether or not we want any linkages between the RSE chapters. If we do, then what should they be? And should they also link to data, code and/or narratives used in the novice chapters?

For instance, the bash shell chapter could use a directory that has text files for the classic english novels in it, the python scripting chapter could then write the `countwords.py` and `collate.py` scripts, the automation chapter could use those scripts in a workflow, and the packaging chapter could package up the scripts for use by others. Or, we could simply decide not to try and maintain links between chapters.

joelostblom:
I would also prefer a linked narrative and think this would be easier for readers to follow. Being familiar with the data would also (I imagine) reduce cognitive load when new concepts are introduced (versus having to learn new concepts and new data). 

I think the story we tell is really important to keep learners engaged and facilitate for them as much as possible (or avoid confusing them, whichever wording you prefer). That is why I opened #226 also, I think we can put more effort into one great narrative if is similar across as many chapters as possible.

As a caveat it seems appropriate to mention that I have not experience in teaching with a disjoint narrative, so don't have any evidence for is leading to a subpar learning experience, maybe it could be refreshing to not see the same data all the time?

elliewix:
Lots of positive discussion here.

We should come up with a coherent starting narrative, then come up with the end game code for each of the chapters.
Once we have that written the group doing the rewrites should go back through the plan and revisit the narrative based on how the programs were actually implemented.
After that we can slice out the relevant concepts from each and start assigning writing tasks.

SaraMati:
[...]

PR #276, opened 2019-11-07
Additions to Python command-line scripting chapter.

gvwilson:


rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
   -h, --help  show this help message and exit
 ```
 
-and issues errors when users give the program invalid arguments:
+As the `--help` hint shows,
+`argparse` can handle [long options][long-option-shell] as well as traditional single-letter options;
+the online help has examples of this and much more.
```
DamienIrving: Should we link to the online help?

rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
...
+
+## How can I configure common options for a program? {#rse-py-scripting-configure}
+
```
DamienIrving: In #156 (and #223) there is some discussion about whether we want to include any content on configuration (i.e. beyond command line options). We never really settled on an answer to that question - this PR basically says that we do want to include it.

rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
...
+A program that does exactly the same thing every time we run it
+isn't as useful as one that can work on different files
+or analyze data with different thresholds.
+Software of all kinds needs to be controlled;
+some things change more often than others,
+so we need a simple, uniform way to specify some options and leave others alone.
+
```
DamienIrving: The paragraph above reads like the sections preceding it never happened. It might be better to start this section by saying that command line options (which were introduced in the previous section) are only one level of / option for configuration. There's also job-specific, user-specific and system-wide configuration files...

rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
...
+4.  Command-line options to change things that commonly change.
+
+This is sometimes called [overlay configuration][overlay-configuration]
+because each level overrides the ones above it:
+the user's configuration file overrides the system settings,
+the job configuration overrides the user's defaults,
+and the command-line options overrides that.
+
+Programmers have invented many formats for configuration files,
+so please do not create your own.
```
DamienIrving: Do we need list/discuss all the different options for configuration files, or should we just cut to the chase and say that YAML is popular so we're going with that?

rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
...
+logfile: "/tmp/log.txt"
+quiet: false
+overwrite: false
+fonts:
+- Verdana
+- Serif
+```
+
+And here's a short Python program that reads and prints that configuration:
+
```
DamienIrving: Should this example yaml code be added to `script_template.py` so that we've got an example script that can be configured via command line options and/or a YAML file?

rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
...
-> by writing Python scripts that can be executed at the command line.
+> We should always include a version number as a field in the dumped configuration
+> (Section \@ref(rse-git-advanced-tag)),
+> and our program should print this when given a `--version` option.
+> We need this because how we interpret options will change over time,
+> and if you don't know what the version of the program was,
+> we'll have to guess what options mean.
+
+## How can I detect errors? {#rse-py-scripting-exceptions}
+
```
DamienIrving: Perhaps error handing should have its own chapter of the book? (Perhaps the very next chapter after scripting?)

rse-py-scripting.Rmd
```diff
@@ -268,10 +717,210 @@ ERROR_MESSAGES[user_language]['cannot_read_file']
...
+but doesn't fill up the disk,
+for example,
+or a centralized logging server of some kind that collates logs from many different systems.
+We don't need any of these when we start,
+but the data engineers and system administrators who eventually have to install and maintain your programs
+will be very grateful that you used `logging`,
+because then they can set it up the way they want with very little work.
+
+## How should I log my program's configuration? {#rse-py-scripting-log-config}
+
```
DamienIrving: How does this section relate to the previous section: "How can I keep a record of the actual configuration that produced particular results? {#s:configure-dump}"

rse-py-scripting.Rmd
```diff
@@ -103,42 +133,461 @@ optional arguments:
...
+```
+
+```shell
+$ cat thesis.yml
+```
 
-FIXME: In this section we will take the code developed in the novice book
-and turn it into a Python script that can be executed at the command line.
-See [this lesson](https://carpentrieslab.github.io/python-aos-lesson/04-cmdline/index.html)
-for an example of this approach in action.
```
DamienIrving: Once we've settled #240 (i.e. the narrative that runs through the RSE book), I'd argue that this placeholder about taking the code we've developed (as part of the narrative) and turning it into a command line program is still relevant.

rse-py-scripting.Rmd
gvwilson: :+1:

rse-py-scripting.Rmd
gvwilson: I re-read Taschuk's paper, and I think we do need some discussion of persistent configuration for reproducibility - she has a good rant about shell scripts that rely on defaults, but then defaults change.

rse-py-scripting.Rmd
gvwilson: Good point - paragraph removed and intro amended.

rse-py-scripting.Rmd
gvwilson: I included it because people are still likely to run into it. I took out the mention of Python data structures as a configuration format.

rse-py-scripting.Rmd
gvwilson: Good catch - example enlarged.

rse-py-scripting.Rmd
gvwilson: I think it would be pretty short to be a chapter on its own. On the other hand, if the Python scripting chapter is re-labelled "How to write a robust command-line software tool", then maybe the config and error handling stuff fits more naturally?

rse-py-scripting.Rmd
gvwilson: Ah - good catch. Section replaced with a short callout box. Sorry about that.

rse-py-scripting.Rmd
gvwilson: :+1:

rse-py-scripting.Rmd
DamienIrving: I guess the question that came to mind was "is error handling an explicitly command line tool thing, or are there other contexts in which it's applicable such as when working more interactively in a notebook?" We can leave it in this chapter for now :-)

gvwilson:
It's definitely not a command-line only thing, but I'm not sure where else to put it - I'll merge this now and we can move it once we have a common thread running through the chapters.

PR #283, opened 2019-11-19
RSE narrative

DamienIrving:
I've started to write down some thought about what the narrative in the RSE book will be and how best to weave it into the chapters. (A conversation that started at #240.)

Feel free to comment and I'll update the document as we go :smile:


gvwilson:
Thanks very much for putting this together - I like this plan.

1. I think "Working in Teams" come after "Advanced Git" - now that you have the tools, here's the social machinery required to use them.

2. Maybe this volume is Python-only and we do a separate (parallel) one for R? Shoehorning both into one book is looking increasingly awkward, and this would parallel the split in the intro material.

lwjohnst86:
Definitely agree with @gvwilson about splitting the content by language, in some chapters there's major differences. Likewise, some chapters are just not necessary from the R perspective. So +100 :grin:

lwjohnst86:
Looks good @DamienIrving. 

As for the practicalities of splitting, we can maybe start moving the content into subfolders? Bookdown allows searching into [sub directories](https://bookdown.org/yihui/bookdown/configuration.html). I can do that for the R stuff...?

rse-narrative.md
```diff
@@ -0,0 +1,157 @@
...
+This chapter can stay pretty much as is.
+The introductory paragraphs about Zipf's Law can be removed 
+(because they should now appear in the introduction chapter)
+and the workflows developed in the chapter could be extended to also include
+`plotcounts.py` and `testfit.py`.
+
+### Chapter 10: Project structure
+
+Now that we've written all our scripts for the Zipf's Law analysis,
+this chapter should structure it all properly (i.e. with license files and README's etc).
```
lwjohnst86: This I think for the R side would be better at the beginning, to begin structuring everything in an R package format.

rse-narrative.md
```diff
@@ -0,0 +1,157 @@
...
+### Chapter 14: Correctness
+
+This chapter introduces testing and could be updated to write tests for the Zipf's Law code.
+For example, `countwords.py` could be tested using an input text file where the count
+for each word is already known.
+
+I think this chapter needs to come before the chapter on continuous integration,
+since it's these tests that Travis CI will implement.
+
+### Chapter 13: Python packaging and Chapter 12: R packaging
```
lwjohnst86: I think that R packaging should be started from the beginning and start building from there (for the R split if we do it). We would have to make sure that we focus strongly on live-coding/exercises/practical side of R package development so we don't start mimicking the [R Packages](http://r-pkgs.had.co.nz/) book. With the inclusion of teams, collaboration, etc, I think we can differentiate ourselves enough.

rse-narrative.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Chapter 15: Publishing
+
+This lesson could consider (hypothetically) where the results, data and code from the Zipf's Law
+analysis could be published.
+
+
+## Outstanding issues
+
+- Where should "Chapter 9: Working in teams" go in the chapter order?
```
lwjohnst86: Would this chapter maybe go before publishing?

rse-narrative.md
DamienIrving: I agree that the R book could have a completely different order for the chapters - I've edited the document so it only makes reference to the Python book.

rse-narrative.md
DamienIrving: I've followed Greg's suggestion and put it after advanced git.

DamienIrving:
@lwjohnst86 I think this is ready to merge so that people can find the `rse-narrative.md` file in the main repo if they are wondering what the plan is going forward.

After you've separated out the Python and R books, I'll submit a PR re-ordering the chapters in the Python book into the proposed new chapter order.

I'll then start work on the code for the final Zipf's Law package. Once that code is complete (i.e. once we know what end point we're aiming for), I'll start assigning people with the task of updating the narrative in specific RSE Python chapters.

How does that sound?

joelostblom:
@DamienIrving Sorry that I'm late to comment here, the new structure looks great! Thanks for reordering it.

My only comment would be that some of the project structure might be revealed when we talk about commands such as ls and grep and when we save files. I guess we could have a short intro to the structure in the beginning and save files in their final place, but then go more in depth into project structure before we talk about python packaging. The alternative would be to save elsewhere initially and then reorganize when we have introduced the project structure. I'm leaning towards the former.

Issue #288, opened 2019-11-26
Discussion: testing fit with log-normal distribution


gvwilson:
1.  We want to add a script `testfit.py` that tests how well data fits the "expected" distribution according to Zipf's Law.
2.  We don't really know what's expected: should it be log-normal?

-   Thread: https://twitter.com/RobertDotKern/status/1198389738752892928
-   Jake Vanderplas article: https://jakevdp.github.io/blog/2017/11/09/exploring-line-lengths-in-python-packages/

cwickham:
**Log Normal v.s Power Law**

* I didn't find any convincing mechanisms that would suggest word frequencies should be Log Normal.  In contrast, there is a reasonable (but unrealistic) mechanism for why word frequencies might be power law distributed.

* When the probability density function is plotted on a log-log scale:
    * it will appear **linear** (at least in the tails), for a power law distributed variable, and
    * it will appear **quadratic**, for a  Log Normally distributed variable.
    
    Therefore, @newman2005power suggests that since a quadratic can be well approximated by a line on small regions of the scale, a log normally distributed variable may well appear power law distributed.  You can probably make the reverse argument, that a line can be well fit by a quadratic (with its extra parameter), and as such, a log normal may well fit as well or better to data that is in fact power law distributed.

* A log normal is a distribution for continuous random variables, but one might argue word frequencies, since they are counts, are discrete.  *This may not matter much for large texts*.

**Recommendations**

* I suggest we focus on fitting the power law, and don't fit a Log Normal at all.  

* Following [Moreno-Sánchez, Font-Clos, and Corral
(2016)](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0147073) we fit the discrete form with the power law on the survival function with a single parameter: the exponent in the power function (the minimum value of frequency for which the power law applies is assumed to be 1).  

* Fit using maximum likelihood following recommendations of [Clauset, Shalizi, and Newman (2009)](https://arxiv.org/abs/0706.1062).

**Questions**

* Do we want to do any goodness of fit testing?  I.e. we can estimate the exponent, but do we want to ask, is a power law a good fit?  From `automate.Rmd` it looks like the answer is, yes.  This will be more involved code: the recommended approach involves a Monte Carlo simulation to get a p-value.

## Notes

[...]

PR #293, opened 2019-12-10
Code for Zipfs Law narrative

DamienIrving:
In this PR I'm developing the code for the Zipfs Law narrative that will run through the RSE books. The scripts that are central to this include:  
- `countwords.py` reads a text file and creates a CSV file with two columns: a word and how many times the word occurs
- `collate.py` takes one or more of these two-column CSV files as input and sums the counts for all occurrences of each word
- `plotcounts.py` creates a plot that shows word rankings on the X axis and word counts on the Y axis  
- `testfit.py` compares actual distributions against theory and give a fitting score

The final two scripts will make use of the background work from #288


k8hertweck:
Any updates on this, @DamienIrving ?

DamienIrving:
@k8hertweck Should have a chance to finish this up in the next week or two (i.e. many days in advance of our next meeting), so that we can start allocating chapters to people for updating.

DamienIrving:
**HELP!** Does anyone have experience with regular expressions? (I have zero experience.)

I'm working on `countwords.py` and the part of the script responsible for extracting all the words from a string (e.g. an entire book like `zipfs-law/data/anne_of_green_gables.txt`) looks like this at the moment:

```
from collections import Counter
from unidecode import unidecode

raw_text = unidecode(raw_text)
word_list = re.findall(r"\w[\w']*(?:-\w+)*", raw_text.lower())
word_counts = Counter(word_list)
```

When finding all the words, I'd like contractions and hyphenated words to count as a single word (e.g. `don't` and `up-to-date`, not `don`,`t`, `up`, `to` and `date`). It also needs to handle the fact that some of the books use a curly (as opposed to straight) apostrophe (I used the `unidecode` library to get around this).

The code as it stands does most of what I want, but I'm getting tripped up with books that use an underscore for emphasis. For instance, I'd like `_not_` to be counted as `not` instead of `_not_`. I also have no idea if the pattern I'm using (`\w[\w']*(?:-\w+)*`) is the most elegant solution. 

Can anyone help with this?

joelostblom:
Advice from another non-expert on regex, apply salt as needed:

My understanding is [`\w` is shorthand for `[a-zA-Z0-9_]`](https://stackoverflow.com/a/11874899/2166823). So, I would substitute `[a-zA-Z0-9]` (or even `[a-zA-Z]` if we don't want numbers) for `\w`. It seems like not every `\w` needs to be replaced, but I would still do it for clarity.

So something like this:

```python
from collections import Counter
import re

raw_text = "some 0 string __of__ **words** _not. Capital of NO*T here .. _ and don't up-to-date _not_ this *time*"
word_list = re.findall(r"[a-zA-Z][a-zA-Z']*(?:-[a-zA-Z]+)*", raw_text.lower())
word_counts = Counter(word_list)
word_counts
```

    ## Out
    Counter({'some': 1,
             'string': 1,
             'of': 2,
             'words': 1,
             'not': 2,
             'capital': 1,
             'no': 1,
             't': 1,
             'here': 1,
             'and': 1,
             "don't": 1,
             'up-to-date': 1,
             'this': 1,
             'time': 1})

If you want to keep `\w` instead, I think you would have to add a step. Either `replace('_', '')` (or more sophisticated only for surrounding underscores with `re.sub()`) on the `raw_text` string, or `strip('_')` for each word in the word list (something like `word_list = [w.strip('_') for w in word_list if w.strip('_') != '']`).

joelostblom:
By the way, if the only reason for using `unidecode` is the curly apostrophes, we could replace them manually instead:

```python
raw_text.replace("’", "'")
```

And so on for all of `‘`, `’`, `“`, and `”`. This might be clearer since it is more explicit, but maybe it is useful to learn about the library for more complicated cases?

joelostblom:
As for something less verbose, how about this?

```python
word_list = re.sub("[^a-zA-Z]", " ",  raw_text.lower()).split()
```

From https://stackoverflow.com/a/6181784/2166823

elliewix:
So here's the problem.

Are you trying to just demonstrate how this can be done or are you trying to create the world's best word finder script?  In general, `\w+` should be completely fine to find words. Yes, it will fail on `don't` or `super-cool`. We could match those easily with `[\w'-]+` but then that will match things that have m-dash stuff like `"Lynde--a"`.  

You can solve some of this with some string manipulation.  For example, replace any series of `-` with white space. Remove `'` so `don't` becomes `dont` and will match like a word. But are you teaching string parsing here?

`(?:\w)(?!.*-{2,})[\w'-]+` will accomplish the desired "multiple dashes but not in a row plus contractions". But this involves about 10 separate regex concepts.

**So here's my question to you:  is it sufficient to explain that `\w+` is imperfect and there will be noise, or is it worth trying to explain the complexity of these more powerful expressions?** (bolded because of wall of text)

Also, using regex compiling is better practice because you get access to more control.

``` python
findwords = re.compile("(?:\w)(?!.*-{2,})[\w'-]+", re.IGNORECASE)

words = re.findall(findwords, text)
```

DamienIrving:
Thanks for your wise words, @elliewix. The point of the lesson is definitely not to teach regex and the final plot looks basically exactly the same whether we use `\w+` or a more sophisticated pattern, so I'll just use `\w+`.

(While the lesson isn't about regex, it's probably useful for people to at least know that regex is a thing, so I think I'll go with `\w+` instead of `text.split()`.)

DamienIrving:
This is ready to review. It would be great to get it merged prior to our meeting on 7 January so at that meeting we can assign chapters to people.

Few comments:  
* The best way to understand this PR is to read [`narrative.md`](https://github.com/DamienIrving/merely-useful.github.io/blob/zipfs_code/py-rse/narrative.md)
* The only thing left to do is to fit a curve to the data in `plotcounts.py`, which is something I'm working on with @cwickham (see #288). I don't think we should hold up the merging of the PR just for that, as it only relates to one chapter
* The checks are failing for the same reason they are in the `book` branch at the moment (see #294)

zipfs-law/bin/plotcounts_branch_reject.py
```diff
@@ -0,0 +1,33 @@
...
+import matplotlib as mpl
+
+
+def main(args):
+    """Run the command line program."""
+
+    input_csv = args.infile if args.infile else sys.stdin
+    df = pd.read_csv(input_csv, header=None, names=('word', 'word frequency'))
+    df['rank'] = df['word frequency'].rank(ascending=False)
+    df['1/rank'] = 1 / df['rank']
```
gvwilson: maybe `inv_rank` instead of `1/rank` as a name unless we're planning to show people later that using special characters like `/` in column names makes things difficult? I.e., you can't use `df.1/rank` but you can use `df.inv_rank`.

zipfs-law/bin/plotcounts_branch_keep.py
```diff
@@ -0,0 +1,37 @@
...
+import argparse
+import pandas as pd
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+
+
+def main(args):
+    """Run the command line program."""
+
+    input_csv = args.infile if args.infile else sys.stdin
```
gvwilson: We need to explain somewhere that `pd.read_csv` takes either a stream or a string as its first argument - I only learned this recently.

zipfs-law/bin/plotcounts.py
```diff
@@ -0,0 +1,107 @@
...
+    df = pd.read_csv(input_csv, header=None, names=('word', 'word frequency'))
+    df['rank'] = df['word frequency'].rank(ascending=False)
+    df.plot.scatter(x='word frequency', y='rank', loglog=True,
+                    figsize=[12, 6], grid=True,
+                    xlim=args.xlim, ylim=args.ylim)
+
+    alpha, beta = get_power_law_params(df['word frequency'].values)
+    print('alpha:', alpha)
+    print('beta:', beta)
+    plot_fit(args.xlim, df['rank'].values[-1], beta)
```
gvwilson: why `values[-1]` ?

zipfs-law/bin/plotcounts_branch_reject.py
DamienIrving: Good suggestion.

zipfs-law/bin/plotcounts_branch_keep.py
DamienIrving: Agree - will definitely need to explain that in the text of the chapter.

zipfs-law/bin/plotcounts.py
DamienIrving: Because the rank column has been sorted, values[-1] represents the maximum rank (or the rank of the words that only appear once). We need that when plotting the fitted line.

More specifically, relationships of the form y = bx<sup>m</sup> appear as a straight line on log-log plots. Taking the logarithm of the equation yields log(y) = mlog(x) + log(b). Setting X = log(x) and Y = log(y), which corresponds to using a log–log graph, yields the equation: Y = m X + c, where m is the slope of the line (in our case m = -&beta; + 1) and c = log(b) is the intercept on the (log y)-axis, meaning where log x = 0, so, reversing the logs, b is the y-value corresponding to x = 1 (the maximum rank).

Issue #294, opened 2019-12-31
Current version of book doesn't compile


DamienIrving:
I get the following error when running `make html` using the HEAD of the `book` branch:

```
$ make html
mkdir -p _book
cp ./_index.html _book/index.html
cp -r climate-data data src zipfs-law _book
rm -f py.Rmd
cp py-index.Rmd index.Rmd
Rscript -e "bookdown::render_book(input='index.Rmd', output_format='bookdown::gitbook', config_file='_py.yml'); warnings()"


processing file: py.Rmd
  |                                                                 |   1%
   inline R code fragments

  |.                                                                |   1%
label: py-index-setup (with options) 
List of 1
 $ include: logi FALSE

  |.                                                                |   2%
  ordinary text without R code


label: novice-goals-setup (with options) 
List of 1
 $ include: logi FALSE

  |..                                                               |   3%
  ordinary text without R code


label: py-intro-setup (with options) 
List of 1
 $ include: logi FALSE

  |..                                                               |   4%
  ordinary text without R code

  |...                                                              |   4%
label: py-getting-started-setup (with options) 
List of 1
 $ include: logi FALSE

  |...                                                              |   5%
  ordinary text without R code

  |....                                                             |   5%
label: unnamed-chunk-1 (with options) 
List of 1
 $ child: chr "questions/py-getting-started.md"



processing file: ./questions/py-getting-started.md
  |.................................................................| 100%
  ordinary text without R code


  |....                                                             |   6%
  ordinary text without R code

  |....                                                             |   7%
label: py-getting-started-spyder (with options) 
List of 2
 $ echo   : logi FALSE
 $ fig.cap: chr "Spyder"

  |.....                                                            |   7%
  ordinary text without R code

  |.....                                                            |   8%
label: unnamed-chunk-2 (with options) 
List of 1
 $ engine: chr "python"


  ordinary text without R code

  |......                                                           |   9%
label: unnamed-chunk-3 (with options) 
List of 1
 $ engine: chr "python"


  ordinary text without R code

  |......                                                           |  10%
label: unnamed-chunk-4 (with options) 
List of 1
 $ engine: chr "python"

  |.......                                                          |  10%
  ordinary text without R code

  |.......                                                          |  11%
label: py-getting-started-short-script (with options) 
List of 2
 $ echo   : logi FALSE
 $ fig.cap: chr "A Short Script"


  ordinary text without R code

  |........                                                         |  12%
[...]

PR #295, opened 2019-12-31
Starting fix for #294 (failure to load dog licensing data).

gvwilson:
Partial fix for #294:

1.  Uncompress `data/nyc-dog-licenses.csv.gz` as needed.
2.  Modify `.gitignore` to ignore the uncompressed file.
3.  Correct the filepath in `py/data-manipulation.Rmd`.

This still does not produce a working build: we get a complaint about subtracting `datetime` objects with and without timezone information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ _bookdown_files
 _main.Rmd
 _novice_py
 _novice_r
+data/nyc-dog-licenses.csv
 birds.png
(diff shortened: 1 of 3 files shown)


lwjohnst86:
@gvwilson I have a pretty simple fix for this that doesn't require uncompressing. I can't get to it in the next few days, but next week I can, if you can wait until then.

PR #296, opened 2019-12-31
Temporary fix for #294: only build R and Python RSE volumes automatically.

gvwilson:
1.  Rename the `rse` target in `Makefile` to `r-rse` to accurately reflect what it does (it builds the R RSE volume, not both of the RSE volumes).
2.  Modify `.travis.yml` to only build the R and Python RSE volumes for now while we try to figure out the broken `datetime` manipulation described in #295.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ py : _book/py/index.html _book/py/rse.pdf
 py-rse : _book/py-rse/index.html _book/py-rse/py-rse.pdf
 
 ##   r-rse      : rebuild RSE R HTML and PDF.
-rse : _book/r-rse/index.html _book/r-rse/r-rse.pdf
+r-rse : _book/r-rse/index.html _book/r-rse/r-rse.pdf
 
 #-------------------------------------------------------------------------------
 


Makefile
```diff
@@ -110,7 +110,7 @@ py : _book/py/index.html _book/py/rse.pdf
 py-rse : _book/py-rse/index.html _book/py-rse/py-rse.pdf
 
 ##   r-rse      : rebuild RSE R HTML and PDF.
-rse : _book/r-rse/index.html _book/r-rse/r-rse.pdf
+r-rse : _book/r-rse/index.html _book/r-rse/r-rse.pdf
```
lwjohnst86: duh, forgot this. Thanks for the catch.

.travis.yml
```diff
@@ -15,7 +15,7 @@ addons:
 
 script:
   - Rscript -e "tinytex::install_tinytex()"
-  - make everything
+  - make r-rse py-rse
```
lwjohnst86: @gvwilson I have a fix as per my comment in #295, if you can revert this change, we can merge this edit it.

gvwilson: Change reverted. Please try to get a fix for the data file in place before Tuesday's meeting - thanks.

Issue #308, opened 2020-01-16
Do we want to make the first argument to `countwords.py` positional instead of controlled by the `--infile` flag?


joelostblom:
I am asking because the automate chapter does not use the flag currently, and before I add it to all the examples I just wanted to check if there is a specific reason we want to use a flag for input files. I am fine with either, but I think it is more common to pass input files as positional arguments: e.g. `python bin/countwords.py data/moby_dick.txt` instead of  `python bin/countwords.py --infile data/moby_dick.txt`.

If the flag is just there to differentiate when `stdin` should be used, I think it is nicer to rewrite the argparser to have `stdin` as the default if no positional argument is used. We just need to change this section

```python
    parser.add_argument('--infile', type=str, default=None,
```

to

```python
    parser.add_argument('infile', type=str, default=sys.stdin, nargs='?',
```

and higher up just change `if args.infile:` to `if args.infile != sys.stdin:`.

I can make this changes unless everyone else has used `--infile` a lot already and would have to rewrite. If we don't plan to use `stdin` with `countwords.py` in any chapter, we can simplify further, but I am not sure if it is there to be taught specifically with pipes somewhere. @DamienIrving You have the best overview, let me if this makes sense or if I am missing something.

joelostblom:
I should add that in `collate.py`, the input files are passed as positional arguments:

```python
    parser.add_argument('infiles', type=str, nargs='*', help='Input file names')
```

lwjohnst86:
I agree to this. +1

DamienIrving:
@joelostblom Great idea regarding `countwords.py`. Like you suggested, the only reason I used `--infile` was that I wanted to be able to use standard input and I hadn't thought of setting `default=sys.stdin`. Do you want to go ahead and make that change to `countwords.py` in #309 ?

With respect to `plotcounts.py`, if we make `infile` a positional argument that means it will have two positional arguments (`infile` and `outfile`). That means the following pipeline would break, right?

```
$ python countwords.py ../data/anne_of_green_gables.txt | python plotcounts.py anne.png 
```

PR #309, opened 2020-01-16
Zipf scripts suggestions

joelostblom:
Some suggestions that popped up when I was working with the scripts. Sorry that I didn't find time to review more thoroughly in your previous PR @DamienIrving. None of these suggestions changes the behavior or usage of the scripts.

@elliewix could you confirm whether it is correct that raw strings are to be used with regex to prevent unexpected Python string processing before the string reaches the regex engine?

### DIFF
--- a/zipfs-law/bin/collate.py
+++ b/zipfs-law/bin/collate.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+"""Combine multiple word count CSV-files into a single cumulative count."""
 
 import sys
 import csv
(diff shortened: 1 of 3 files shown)


joelostblom:
Btw, why do we not autoscale the axes? Is it to show how they can be set? Should we change the default to auto? This is the result when running the make pipeline with all files and the default axes values:

image

DamienIrving:
Actually, before we merge we should resolve #308 ...

elliewix:
The use of raw strings allows for `\` to behave as expected without having to do weird string stuff. This is discussed in the documentation: https://docs.python.org/3/library/re.html 

So what you said is right, but this is a nice example of showing documentation as well.

PR #317, opened 2020-01-21
Revision to correctness chapter with more exercises

### LINKED ISSUE #302: Correctness exercises
Several exercises subheadings exist with a few questions/tasks to reinforce concepts. Further development in the [Exercises subsection](https://github.com/merely-useful/merely-useful.github.io/blob/227dd84db1e7326d6612ce934a90ae5181a76567/py-rse/correct.Rmd#L1461-L1519) needed.

lwjohnst86:
Resolves #302 

I'm not a python person, so I've added several TODO items that need to be addressed for those who are good in python. There are also a few TODO items I left in because they depend on what happens in earlier chapters. I'm also not sure about some of the content, since depending on how earlier chapters go, they may need to be updated to match them.

Basic summary of changes:

- Added 5 more exercises
- Greatly reduced content on floating point values since it felt more of a tangent than related to our stated narrative
- Rearranged some sections to make the overall flow better
- Did general revisions and editing for clarity and concision


py-rse/correct.Rmd
```diff
@@ -633,471 +469,51 @@ def test_text_must_not_be_empty():
...
 In Python,
 we can use `pytest.approx`,
 which works on lists, sets, arrays, and other collections,
 and can be given either relative or absolute error bounds.
 To show how it works,
 here's an example with an unrealistically tight absolute bound:
 
-```python
+```{python}
 from pytest import approx
```
joelostblom: I didn't know about `approx`, I have been using `np.isclose`, guessing `approx` is better for the job but `isclose` has portability so might be worth a mention?

py-rse/correct.Rmd
```diff
@@ -467,51 +366,22 @@ Each test function also has a stereotypical structure:
...
-```
-
 Using helper functions to create fixtures like this
 reduces the [cognitive load][cognitive-load] on whoever has to maintain the tests,
 since they only have to form a [mental model][mental-model] of a few cases
-rather than a new one for each test.
+rather than a new one for each test. 
+Below is an example structure of how a Python helper function may look like:
+
+TODO: Include example Python code of a helper function for fixtures.
```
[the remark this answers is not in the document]
joelostblom: [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) might also deserve a mention here somewhere. It facilitates setting up additional combinations of data to be tested so could be particularly useful with more complex data structures such as data frames.

py-rse/correct.Rmd
```diff
@@ -39,40 +25,28 @@ def numSign(x):
...
-A simple way to test the function interactively in either language
-is to check that it produces the correct value for a few representative cases:
+The simplest way to test the function would be to interactively check it produces the correct values.
+Type out these commands to see that we get what we expect.
 
-```{r manual-test-r-2}
+```{python manual-test-2}
 numSign(2) == 1
 numSign(0) == 0
 numSign(-4) == -1
```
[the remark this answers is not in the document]
lwjohnst86: Yea, sure. Damn didn't even catch that ...duurrrrrr

py-rse/correct.Rmd
lwjohnst86: Wouldn't `isclose` be *less* portable because it relies on numpy? While approx relies on pytest which we use anyway for unit testing?

DamienIrving:
At the moment the narrative about Zipf's law and the code we've developed in earlier chapters (i.e. `countwords.py`, `collate.py` and `plotcounts.py`) doesn't appear in this chapter until the exercises (the first exercise asks the reader to create a simple unit test for `countwords.py`).

Since we need to develop some tests for our code in this chapter that can be implemented using Travis CI in the following chapter, I'm thinking we shouldn't leave the development of the tests for our code until the exercises. For instance, the section on synthetic data could be rewritten so that it develops a test for `countwords.py` by creating a synthetic ebook/text where the count for each word is already known.

In other words, I feel like the focus of this chapter should be writing tests for the existing code that we've developed, rather than coming up with lots of new data analysis examples in order to demonstrate the principles of testing (which is what this chapter does at the moment).

lwjohnst86:
Definitely agree @DamienIrving!! Though in this case it would help to have an idea of how a function within a script gets tested...? Or can a script be tested? Testing would make a lot of sense *after* the packaging chapter, what do you think?

DamienIrving:
@lwjohnst86 You raise a good point regarding the testing of functions versus scripts and how that relates to the packaging chapter.

When we package and release our pip installable zipfs-law package, will that package include command line scripts or just a module/s with all the useful functions we developed for Zipf's law analysis? My sense is it's not common for python packages to come with scripts (and I don't even know if it's possible)?

py-rse/correct.Rmd
joelostblom: That's one way to see it! I mean more portable across testing suites, since it is not relying on pytest specifics

joelostblom:
I agree it would be great if tests could be written for the existing code and only minimal code used to introduce the principle of a test. We don't have to aim for full coverage but can just pick the easiest functions to illustrate the points on a real world example. 

I'm not great at testing theory, but from what I have seen, it is usually individual functions rather than scripts that are being tested. I think we can have testing either before or after making a package, as long as all the functions are in place tests. It might be more natural to do the testing before creating a package (although I can see arguments for either side as long as the package is not released without a testing suit).

You can install command line scripts together with a Python module by setting `entry_points` in `setup.py`, e.g. `entry_points = {'console_scripts': ['<my_new_command>=<folder_name>:<script_name>']}`.

lwjohnst86:
@DamienIrving and @joelostblom think I could merge this in for now and deal with the other details after we've finished earlier chapters (especially the packaging)?=

joelostblom:
And sorry for the late reply.

PR #318, opened 2020-01-22
Additional zipf scripts suggestions

### LINKED ISSUE #308: Do we want to make the first argument to `countwords.py` positional instead of controlled by the `--infile` flag?

joelostblom:
Mostly internal fixes, the only outwards facing change is the one we talked about in #308. I suggest giving a default outfile name to solve that issue, so this will close #308 .

The logic for getting stdin as defaut while still handling files, need an additional tweak so that files have their contents read in rather than their file name. If this is confusing we can go back to handling the logic with if statements in the functions. I think the only downside of the new approach is that it might not be suitable for big programs that requires stdin to continue to be open, but we're note teaching that and otherwise I find it a bit cleaner but maybe requires a bit more explaining (e.g. '-' means stdin as default).

Other that that I think the changes should be well explained in the commit messages, let me know if something is unclear.


joelostblom:
Thanks @DamienIrving! I made a minor change to the default output file location to accommodate running this script from a non-standard folder structure. Merging

PR #326, opened 2020-02-06
change bash basics to zipfslaw

SaraMati:


DamienIrving:
@SaraMati - Is this ready for review? If so, I'm happy to do it.

SaraMati:
@DamienIrving yes, please and thank you.
 and I have a few questions: 
1) how do we share data with the book, and what is the organisation? there are parts that we show the output of `ls` to zipfs-law folder, and for now, I show what is there (including Makefiles etc.) but I assume this is not the folder that is shared beside the book. so I have to change it when I know what will be inside zipfs-law folder. 

2) related to (1): at the end of advanced bash, for teaching `find`, the 'docs' folder has been used as an example, which I think it much better kept the way it is because zipfs-law folder as it is now doesn't have the different subdirectories and file formats to explore the features of `find`. a suggestion is to have the 'docs' folder inside zipfs-law folder. or let me know if this is already have been planned and I've missed it.

3) couldn't find py-rse glossary to fill in the TDLR pages: https://tldr.sh/; the same for 'regular expression' and all the figures (e.g .figure 2.4)

py-rse/bash-basics.Rmd
```diff
@@ -866,100 +862,107 @@ This options works the same way with `mv` and `cp`.
 
 ## How can I run commands on lots of files at once? {#py-rse-bash-basicswildcard}
 
-The `cleaned` directory in our climate data project contains
-precipitation and temperature records from four weather stations in Australia:
+The `data` directory in our zipfs-law project contains
+the text files for several ebooks from Project Gutenberg:
```
DamienIrving: This can be [Project Gutenberg][project-gutenberg] to link to the glossary entry

py-rse/bash-basics.Rmd
```diff
@@ -866,100 +862,107 @@ This options works the same way with `mv` and `cp`.
...
+data/ethan_frome.txt            data/sherlock_holmes.txt
+data/frankenstein.txt           data/time_machine.txt
+data/jane_eyre.txt              data/treasure_island.txt
 ```
 
-while `cleaned/b*.csv` only matches the four whose names begin with a 'b':
+while `data/c*.txt` only matches the four whose names begin with a 'b':
 
 ```shell
 $ ls cleaned/b*.csv
```
DamienIrving: I think you meant `$ ls data/c*.csv`, not `$ ls cleaned/b*.csv`

py-rse/bash-basics.Rmd
```diff
@@ -866,100 +862,107 @@ This options works the same way with `mv` and `cp`.
...
 ```
 
 ```text
-cleaned/badingarra_prec.csv      cleaned/badingarra_temp.csv      cleaned/bellambi_prec.csv        cleaned/bellambi_temp.csv
+data/common_sense.txt  data/count_of_monte_cristo.txt
```
DamienIrving: I think you meant `$ ls data/c*.csv`, not `$ ls cleaned/b*.csv`

py-rse/bash-basics.Rmd
```diff
@@ -1292,33 +1329,33 @@ it reads the output of the upstream command.
...
 we can add a call to the `tail` command:
 
 ```shell
-$ wc -l *.csv | sort -n | head -n 3 | tail -n 1
+$ wc -l *.txt | sort -n | head -n 3 | tail -n 1
 ```
 
 ```text
-    8680 tuggeranong_temp.csv
+  4102 life_of_frederick_douglass.txt
```
DamienIrving: It's a little weird that the final result we save is the name of the third shortest book, because in the text earlier on we ask the question: "which of these books is shortest?"

Perhaps it's better not to introduce the final call to `tail` and instead have the final thing be:
`$ wc -l *.txt | sort -n | head -n 1 > shortest.txt` 
(i.e. the shortest book)

py-rse/bash-basics.Rmd
```diff
@@ -1689,24 +1790,22 @@ $ history
...
 
 ```text
-  ...
-  571  wc -l *.csv | sort -n | head n 3
-  572  wc -l *.csv | sort -n | head -n 3
-  573  wc -l *.csv | sort -n | head -n 3 | tail -n 1
-  ...
+  551  wc -l *.txt | sort -n
+  552  wc -l *.txt | sort -n | head -n 3
+  553  wc -l *.txt | sort -n | head -n 3 | tail -n 1
```
DamienIrving: If you agree with my earlier comment regarding running `wc -l *.txt | sort -n | head -n 1` instead of command 552 and 553 above, then this output from `history` will need to be updated.

py-rse/bash-basics.Rmd
```diff
@@ -1743,54 +1842,55 @@ though,
...
+$ cp data/m*.txt backup
 ```
 
-But what if we want the backup files to have the [extension][filename-extension] `.bak` instead of `.csv`?
+But what if we want the backup files to have the [extension][filename-extension] `.bak` instead of `.txt`?
 `cp` can do this for a single file:
 
 ```shell
-$ cp cleaned/andamooka_prec.csv backup/andamooka_prec.bak
+$ cp data/moby_dick.txt backup/moby_dick.txt
```
DamienIrving: I think this is supposed to be `$ cp data/moby_dick.txt backup/moby_dick.bak`

py-rse/bash-basics.Rmd
```diff
@@ -1743,54 +1842,55 @@ though,
...
        cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
 ```
 
-`backup/*_prec.bak` doesn't match anything—those files don't yet exist—so
+`backup/*.bak` doesn't match anything—those files don't yet exist—so
 what we are actually asking `cp` to do is:
 
 ```shell
-$ cp cleaned/andamooka_prec.csv cleaned/badingarra_prec.csv cleaned/bellambi_prec.csv cleaned/tuggeranong_prec.csv
+$ cp moby_dick.txt mysterious_affair_at_styles.txt
```
DamienIrving: This command would work - it would simply replace the contents of `mysterious_affair_at_styles.txt` with those in `moby_dick.txt`. I think this section was originally written for a directory with more than two files starting with m. There aren't more than two books starting with any letter, so this section might need a minor re-think.

py-rse/bash-advanced.Rmd
```diff
@@ -81,94 +80,91 @@ For example,
...
 What we really want is a way to get the years from any of our files.
-Let's edit `years.sh` again and replace `cleaned/bellambi_temp.csv`
+Let's edit `book_summary.sh` again and replace `data/moby_dick.csv`
 with a special variable `$1`.
 Once our change is made,
 `years.sh` should contain:
 
 ```text
-cut -d , -f 2 $1 | uniq
+head -n 19 $1 | tail -n 11 | grep Author
```
DamienIrving: There are some references to `years.sh` in the paragraph above that need to be removed.

SaraMati:
thanks, I'll apply them today

py-rse/bash-basics.Rmd
SaraMati: true, although I thought the purpose is to show the pipe can be any length (multiple commands), but I guess they get the message with two commands as well, you're right, it's better than suddenly selecting the third shortest book!

SaraMati:
@DamienIrving ready for second review

DamienIrving:
Tagging @lwjohnst86, @gvwilson to figure out why this check is failing.

PR #333, opened 2020-03-05
Avoid log of 0 when beta is 1

joelostblom:
Let me know if there are more elegant solutions. We would need to update this in the lesson where this scripts is created also.


DamienIrving:
That seems like a good solution to [the issue](https://github.com/merely-useful/merely-useful.github.io/issues/288#issuecomment-595005839) you've raised.

@cwickham - what do you think?

cwickham:
Looks like a good solution to me!  I

PR #335, opened 2020-03-05
Testing to see if Travis builds other pdfs

lwjohnst86:
Trying to figure out why Travis fails but my local build is fine...


lwjohnst86:
I really can't figure out why the LaTeX build is failing... there is a warning about TeXLive "freezing" in prep for a new release of TeXLive, so it could be due to that. Whatever the reason, I'm deciding that the PDF version is not necessary right now and we can figure it out hopefully after TeXLive has upgraded...

PR #307, opened 2020-01-16
Zipf changes for the automation chapter

joelostblom:
Mostly minor edits, ready for review.

PR #306, opened 2020-01-15
Working narrative into py-rse/git-cmdline.Rmd

cwickham:
As part of: https://github.com/merely-useful/merely-useful.github.io/issues/240#issuecomment-571508049

PR #338, opened 2020-03-06
Reduce novel texts

joelostblom:
This will eventually contain all the updates related to reducing the number of novel texts. I am also including some other fixes that I notice as I go through the chapters to update the files names (I could open a separate PR for these if you want, but they are mostly at least somewhat related to sections where the other updates are so it felt natural to include them here.

Adding the command I used to search through files in case I need to do this again far in the future when it is gone from my history...:

```zsh
# cd to py-rse first

for f in ../zipfs-law/data.old/*.txt; do
f=${f##*/}
echo "========================================================="
echo $f
echo "========================================================="
rg $f -C 2
echo "\n\n"
done
```


py-rse/bash-basics.Rmd
```diff
@@ -1847,65 +1810,69 @@ $ for filename in s*.txt; do echo $filename; head -n 19 $filename | tail -n 11;
...
 
 ```shell
-$ cp data/*of*.txt backup/*of*.bak
+$ cp data/*e.txt backup/*e.txt.bak
 ```
 
 ```text
-usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
-       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
+cp: target 'backup/*e.txt.bak' is not a directory
```
joelostblom: This is the error I see with with cp 8.31, bash 5.016(1)

py-rse/bash-basics.Rmd
```diff
@@ -870,15 +870,10 @@ $ ls data
...
-anne_of_green_gables.txt        moby_dick.txt
-common_sense.txt                mysterious_affair_at_styles.txt
-count_of_monte_cristo.txt       pride_and_prejudice.txt
-dracula.txt                     README.md
-emma.txt                        sense_and_sensibility.txt
-ethan_frome.txt                 sherlock_holmes.txt
-frankenstein.txt                time_machine.txt
-jane_eyre.txt                   treasure_island.txt
-life_of_frederick_douglass.txt
+README.md         moby_dick.txt
```
joelostblom: Whether upper case names appear first depends on what `LS_COLLATE` is set to. It at least used to be the default on Ubuntu that upper case is before lower case, but I haven't installed that distro in a while so I am not sure about the current state of things. I tested on a friend's mac and there upper case went first. Github sorts upper case first as well.

py-rse/bash-basics.Rmd
```diff
@@ -1847,65 +1810,69 @@ $ for filename in s*.txt; do echo $filename; head -n 19 $filename | tail -n 11;
...
-it reports an error.
+since `backup/*e.bak` is not,
+`cp` reports an error.
 
 Instead,
-let's copy all the files and then rename them:
+let's use a loop to copy files to the backup directory
+and append the `.bak` suffix:
 
 ```shell
```
joelostblom: I could be convinced either way here. The reason I changed this loop was that I have a slight aversion to teach copying twice since this is inefficient and not a habit learners should pick up. Ideally, we would no need `cd`, but I can't think of a way to remove it that does not contain substring matching to exclude the path prefix, which I think is beyond what we teach here.

joelostblom:
I rebased this on my currently open PR because I am building on the changes made there for automate.Rmd

joelostblom:
Rebased on the book branch after merging #307

joelostblom:
@DamienIrving I remember you mentioned planning to write an intro to the zipf's law section, but I couldn't find this. Below are the section we took out from automate.Rmd, with the updated numbers for the remaining novels, in case you have a file somewhere to add this to (or if we don't need this anymore that is also fine):

```

| Book                            | Length (words) |
| ------------------------------- | -------------: |
| dracula.txt                     |         164424 |
| frankenstein.txt                |          78098 |
| jane_eyre.txt                   |         188455 |
| moby_dick.txt                   |         215830 |
| sense_and_sensibility.txt       |         121590 |
| sherlock_holmes.txt             |         107533 |
| time_machine.txt                |          35524 |

The most common words in this [corpus][corpus] appear this many times:

| Word | Count |
| ---- | ----: |
| the  | 44770 |
| and  | 28894 |
| to   | 25264 |
| of   | 24363 |
| I    | 23371 |
| a    | 18933 |
| in   | 14606 |
| was  | 11464 |
| that | 10828 |
| it   | 9820  |


The frequencies don't match Zipf's predictions exactly—for example,
we would expect about 29,000 occurrences of "and"—but
there certainly seems to be a decay curve of some kind.

```

joelostblom:
This PR is read for review.

@DamienIrving @gvwilson @SaraMati I tag you since you have been involved in the chapters I made changes to, please review or let me know if you don't have time and I will tag someone else.

joelostblom:
This PR is read for review.

@DamienIrving @gvwilson @SaraMati I tag you since you have been involved in the chapters I made changes to, please review or let me know if you don't have time and I will tag someone else.

cwickham:
@joelostblom Is this now the list of texts we plan to include?

> | Book                            | Length (words) |
> | ------------------------------- | -------------: |
> | dracula.txt                     |         164424 |
> | frankenstein.txt                |          78098 |
> | jane_eyre.txt                   |         188455 |
> | moby_dick.txt                   |         215830 |
> | sense_and_sensibility.txt       |         121590 |
> | sherlock_holmes.txt             |         107533 |
> | time_machine.txt                |          35524 |

If yes, I'll go ahead and make [git command line](#306) is consistent with this as I make other suggested changes.

joelostblom:
@cwickham Yes that is my suggestion at least. I picked them based on not having to make too many changes to the existing files but I did not go through the open PRs. So if you have a novel you have used extensively, we could switch it for one of the ones in this list that is not used that much, e.g. dracula.

DamienIrving:
@joelostblom The intro to Zipf's Law is in the "motivating example" of https://github.com/merely-useful/merely-useful.github.io/blob/book/py-rse-index.Rmd

joelostblom:
@DamienIrving Great, do you think we need the tables I posted in there? I am not sure how much they add... Actually I think this would be a good addition to one of the bash topics using `wc` I will put it in somewhere if it fits.

DamienIrving:
@joelostblom I don't think we need the tables in the intro :smile:

joelostblom:
@DamienIrving My bad, I searched through the repo for emma.txt, but didn't read through every lecture in detail. I will add emma back in and remove another one, maybe time-machine (although it sounds so interesting =) )

joelostblom:
@DamienIrving I ended up going through the entire py-scripting.Rmd and made several suggestion. Could you review my latest commits? Sorry for not making time for this when you had your PR under review. A couple of questions:

Do you think we should we read the file(s) as a stream in collate.py the same way we are doing it in countwords? Or is there a special reason we are using `open` here instead (multiple files not working otherwise?)?

I haven't changed emma.txt yet, first I wonder why the plotting is different in the lesson versus the script. When you show how to plot `emma.txt` in `py-scripting.Rmd`, you plot over the inverse rank:

```python
df['inverse_rank'] = 1 / df['rank']
df.plot.scatter(x='word_frequency', y='inverse_rank',
                figsize=[12, 6], grid=True)
```

In `plotcounts.py`, we plot over the rank. What is the reason for not doing this the same in both places? Are there instructions elsewhere to change the script to how it looks in `plotcounts.py`? Sorry if I am missing something here.

```python
df['rank'] = df['word_frequency'].rank(ascending=False)
ax = df.plot.scatter(x='word_frequency', y='rank', loglog=True,
                         figsize=[12, 6], grid=True, xlim=args.xlim)
```

py-rse/py-scripting.Rmd
```diff
@@ -18,41 +18,89 @@ but that scaffolding stays more or less the same as programs get larger.
...
 ```python
-import utilities
+print(__name__)
+```
+
+When we run this file directly,
+it will print `__main__`:
+
+```shell
+$ python print_name.py
```
DamienIrving: Do you want to show the output from running `python print_name.py`?

py-rse/py-scripting.Rmd
```diff
@@ -18,41 +18,89 @@ but that scaffolding stays more or less the same as programs get larger.
...
+    # code goes here
+
+
+if __name__ == "__main__":
+    main()
+```
+
+
+> **Quickly running short Python statements**
+>
```
DamienIrving: I'd argue that while this is a neat trick, people don't need to know about it.

joelostblom:
Thanks for the quick review and explanations @DamienIrving! 

1. Yes, that is what I had in mind, moslty for consistency. But I agree with your reasoning of showing different usage patterns. I will make it explicit in the text why we do so.

2. Ah ok, makes sense. I will create the replacement plot accordingly.

py-rse/py-scripting.Rmd
joelostblom: I'm good either way, let's see if someone else has strong feelings.

DamienIrving: Cool. You could merge with it in for now and see what people think in future edits.

cwickham:
@joelostblom @DamienIrving Looks like the scripting chapter is mostly swapping out `emma` for `jane_eyre`.  Can we make sure this is also done in the final mission critical exercise so that I can rely on the existence of `results/jane_eyre.png` in the git lesson?

joelostblom:
@cwickham Yup, I just updated with this change.

@SaraMati @gvwilson Let me know if you have any comments on the bash sections, if not I will go ahead and merge.

lwjohnst86:
To keep things moving, I'll merge this in.

Issue #339, opened 2020-03-09
Discussion: Teach what the shebang line is?


joelostblom:
Or remove its [mention in style.Rmd](https://github.com/merely-useful/merely-useful.github.io/blob/book/py-rse/style.Rmd#L179)?

DamienIrving:
I left the shebang out of the python scripting lesson because from googling around it seemed that it doesn't work / needs to be slightly different depending on the operating system?

I'd be open to adding a call out box in the python scripting lesson to briefly discuss what the shebang is (in case people come across it elsewhere), but unless we can be sure that it will work on any operating system I don't think we should actually add it to our `countwords.py`, `collate.py` and `plotcounts.py` scripts.

joelostblom:
My understanding is that it should be pretty portable nowadays and envs should be fine as long as `/usr/bin/env python` is used (https://docs.python.org/3/using/windows.html#shebang-lines), but I don't have extensive experience running this on different OSes. I agree with your idea of a mention.

gvwilson:
I think we need to teach this - people are going to see it a lot when they start working with other people's scripts.

joelostblom:
Good, I will PR when the current PRs have closed since there is a lot of new material under review currently. Also mentioned in #341

joelostblom:
Decided on the meeting today that we will add back shebangs. @DamienIrving I can try this first in packaging and if it works well there, you can add it back to scripting?

DamienIrving:
@joelostblom That sounds like a good plan :smile:

PR #341, opened 2020-03-11
Package section zipf updates

joelostblom:
This is mostly reordering of sections, removal of old code, and rewriting of the text to reflect that. Before I recreate the code section with our package, there are several things we need to decide on:

1. Which virtual environment do we teach? My understanding is that `virtualenv` is not used that often, and that `pyenv` and `conda` are more popular. If we are already teaching how to install anaconda, I vote for also using `conda` for virtual envs.

2. There are more modern and structured approaches to install packages than using setuptools direclty. I am mostly thinking of using a [pyproject.toml](https://www.python.org/dev/peps/pep-0518/) file via [poetry](https://python-poetry.org/) or [flit](https://flit.readthedocs.io/en/latest/). I haven't used this myself much, but it does seem pretty neat and more standardized. Do we want to get into this more than just a mention?

3. This chapter currently uses "Terry Pratchett" in file paths and author sections. Should I just replace with "username" or do we have a specific name we have been using throughout?

4. Change README.rst to README.md? Sphinx supports .md, are we ok with this change? I don't know which, if any, caveats there are to using .md for *simple* sphinx documentation. Although ReST is the canonical Python documentation markup, I think it could be somewhat daunting to feel like you need to learn a new language to write documentation. It might therefore be worthwhile mentioning that sphinx works with markdown documents even if it doesn't support directives and roles to the same extent (I believe), to lower the threshold for getting started, especially for simpler projects markdown can be completely sufficient and is also supported at RTD.

    - https://stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rst
    - https://www.sphinx-doc.org/en/master/usage/markdown.html
    - https://docs.readthedocs.io/en/latest/intro/getting-started-with-sphinx.html#using-markdown-with-sphinx
    - There is also MKdocs, but now that sphinx supports markdown I don't feel like it is necessary to go through, not sure if it is worth briefly mentioning (although I am not an expert on the differences versus sphinx with markdown).

5. We are using shebang lines for the command line scripts that will serve as the entry points when installing the package. I don't know if these will work without shebangs somehow (I doubt it), so we might want to spend some time introducing them (#339).

6. There are different structures possible for packages, please review and see if you agree with what I suggest. If so, should we develop in the package directory from the get go instead of in `bin/` (lots of changes to other files)? Or should I copy all the scripts into a new directory inside or outside the bin dir? Essentially this ties back to how to think of scripts, are they throwaway things that can be kept in bin or src, or always part of a package (just not always formalized) so we should always name the dirs as per Python conventions (the package's name, or potentially src/name) (in general I think `src/` is more common than `bin/`, so that's another (minor) reason to consider changing).

### DIFF
--- a/py-rse/package-py.Rmd
+++ b/py-rse/package-py.Rmd
@@ -902,7 +902,7 @@ to find, understand, and install our package.
 
 ## What does it mean to install a Python package? {#py-rse-package-py-install}
 
-We have made made and published a Python
+We have made and published a Python
 package that anyone with an internet connection can install.
 But what actually happens when you install a Python package?
 
@@ -946,8 +946,8 @@ always looks there first -
 and the rest are system paths for
 the Python installation you are using.
 
-To install a package simply means to copy the contents of the
-package in some form into one of the directories that Python will search in.
+Installing a package means that the contents of the package are copied
+into one of the directories that Python will search in.
 Most commonly when installing packages,
 they will be copied into the `site-packages` directory.
 This means that you can "install" packages
@@ -1076,7 +1076,7 @@ We'll try that here:
 
 1. Add the lines `from .generate import *` and `from .check import *` to
 `zipfpy/__init__.py`.
-2. In the test suite, you can now just call `import zipfpy` and refer to,
+2. In the test suite, you can now call `import zipfpy` and refer to,
 _e.g._, `zipfpy.make_zipf(...)` instead of `zipfpy.generate_make_zipf(...)`.
 Make those changes
 3. Run the test suite with `pytest`
--- a/py-rse/package-py.Rmd
+++ b/py-rse/package-py.Rmd
@@ -252,116 +252,15 @@ using virtual environment names that match the names of your projects
 (and branches, if you're testing different environments on different branches)
 quickly becomes essential.
 
+### How can I install my package? {#py-rse-package-py-install}
 
-## What does it mean to install a Python package? {#py-rse-package-py-install}
-
-By the end of this chapter, you'll have made and published a Python
-package that anyone with an internet connection can install.  You've
-certainly installed Python packages before, such as with `pip install`;
-the `pip` tool is the most common way to install Python packages.
-The command <code>pip install <em>package</em></code>
-checks to see if the package is already installed (or needs to be upgraded);
-if so,
-it downloads the package from [PyPI][pypi] (the Python Package Index),
-unpacks it,
-and installs it.
-
-But what does it mean to install a Python package?
-
-You've seen that when you have a Python module or package in the current
-directory named, say, `zipfpy`, Python will successfully find that module
-and directory when it runs `import zipfpy`.  But clearly having all of the
-Python standard library, all installed packages, and our own code in the
-same directory gets unwieldly quickly - especially if we want to run our
-routines from a directory full of data!
-
-Just as the `PATH` environment in the shell contains a list of directories that
-the shell searches for programs it can execute, in Python a variable `sys.path`
-(in the system-standard `sys` package) contains a list of the directories to
-search.  Create a Python file `showpath.py` containing the following, and
-then run it with `python showpath.py`:
-
-```python
-import sys
-
-for directory in sys.path:
-    print(directory)
-```
-
-If you are using the Anaconda distribution of Python and you run this script
-from your desktop, you will see a list that looks like this:
-
-```
-/Users/tpratchett/Desktop
-/Users/tpratchett/anaconda3/lib/python36.zip
-/Users/tpratchett/anaconda3/lib/python3.6
-/Users/tpratchett/anaconda3/lib/python3.6/lib-dynload
-/Users/tpratchett/anaconda3/lib/python3.6/site-packages
-/Users/tpratchett/anaconda3/lib/python3.6/site-packages/aeosa
-```
-
-the first will be your current directory - a Python environment
-always looks there first - and the rest are system paths for
-the Python installation you are using.
-
-To install a package, then means to copy the contents of the
-package in some form into one of the directories that Python
-will search from.
-
-## How can I create an installable Python package? {#py-rse-package-py-package}
-
-People can always get your package by cloning your repository and copying files from that
-(assuming your repository is accessible,
-which is should be for published research),
-but it's much friendlier to create something they can install
-(and it will be easier for us when it comes to running test suites
-and distributing, too!)
-
-For historical reasons,
-Python has several ways to build an installable package.
-We will show how to use [setuptools][setuptools],
-which is the lowest common denominator;
-[conda][conda] is a modern does-everything solution,
-but the setuptools approach we describe below will
-allow everyone, regardless of what python distribution
-they use, to easily use your package.
-
-To use `setuptools`,
-we must create a file called `setup.py` in the directory *above* the root directory of the package:
-
-```text
-+- setup.py
-+- test_zipfpy.py
-+- zipfpy
-    +- __init__.py
-    +- check.py
-    +- generate.py
-```
-
-The file `setup.py` must have exactly that name,
-and must contain these lines:
-
-```python
-from setuptools import setup
-
-setup(
-    name='zipfpy',
-    version='0.1',
-    author='Greg Wilson',
-    packages=['zipfpy']
-)
-```
-
-This is enough to be able to install the package from the
-local copy source code using standard tools!
-
-Let's create a virtual environment to test the installation:
+Let's proceed to install our package inside this virtual environment
 
 ```text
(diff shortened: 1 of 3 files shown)


py-rse/package-py.Rmd
```diff
@@ -20,445 +20,110 @@ This lesson shows you how to use Python's tools to create and share libraries of
...
-A package in Python is no more than a directory containing
-Python code, and by convention a Python file with a special
-name, `__init__.py` (this was a requirement before Python 3.3).
+```text
+zipfs_law
+├── zipfs_law
+│   ├── __init__.py
+│   ├── collate.py
+│   ├── countwords.py
+│   ├── mymodule.py
```
DamienIrving: It occurs to me that if we are packaging `zipfs_law` for release then we probably want a better / more realistic name for `mymodule.py`. Perhaps `csv_tools.py`?

gvwilson:
:+1: to using `setuptools`, for the same reason that we're teaching Make: the newer tools are probably better, but their user base is a small fraction of that of the older, more established tools.

Also :+1: on `conda` - that's one that _has_ managed to displace an older tool.

I'll be sad to see our mentions of Pterry go, but c'est la vie…

joelostblom:
1. conda it is.
2. I agree that it is often important to know the default minimal way of doing something. I also think that there is reason to mention approaches that tries to address the shortcomings of the defaults, especially when it is not only about conveniences but also robustness. The motivation of pep518 that I linked above goes into some of these in more detail. I guess I am trying to say that I am still split here, but will keep setuptools for now and simply mention pyproject.toml (also because I am short of time do rerwrite and read more on the pros and cons of each approach).
3. Agree, I will use `/home/amira`.
4. The [sphinx](https://www.sphinx-doc.org/en/master/usage/markdown.html) and [readthedocs](https://docs.readthedocs.io/en/latest/intro/getting-started-with-sphinx.html#using-markdown-with-sphinx) documentation also mention how to use markdown.
    - One important advantage of using markdown is that there is no hurdle to get started writing documentation. And since documentation is already something that seems hard to get people to write in the first place, adding extra obstacles seems less than ideal.
        - Given this, I am going to make a proposal: How about we focus on how to write a great readme that includes documentation? And then include the current sphinx/rmd section for where to go once your package outgrow a single page readme? I know a single page is not as ideal as a well structured wiki for documenting bigger projects, but I think it is more than sufficient for projects of the scale that we describe here and it is infinitely better than no docs at all.
            - Best practices would include using a detailed ToC, proper formatting, examples, installation instructions.
5. I also think it is good to teach this.
6. I think the line between messing around and developing a package doesn't have to be strict. Why not mess around in a proper directory structure? We have already put the data in a `data/` dir, so instead of developing code in `zipfs-law/bin/`, we could just develop in `zipfs-law/zipfs_law` or `zipfs-law/src/zipfs_law` and the code would be ready to be packaged. The advantage of this is also that the import statements in any notebooks or scripts using this code does not have to change just because we turn it into a package.

joelostblom:
On that last point, this is a comment I made on a previous PR:

There is an advantage of either naming src the same name as the root (e..g `my_project`, the convention for python packages), or having a folder within src named `my_project`. Putting `__init__`.py within this folder means that when you install this as an editable package in your environment, your functions can now be imported as `my_project.module.function` vs `src.module.function`. The latter would cause name collisions if functions are shared between projects, while the former makes it clear where the functions are from. 

There is a ticket discussion aroudn this on the DS cookie cutter page, and it is also the way of [py-scaffold](https://github.com/pyscaffold/pyscaffoldext-dsproject) (which is a (self-proclaimed) enhancement of DS cookie cutter) with a slightly less complicated folder structure)

py-rse/package-py.Rmd
```diff
@@ -1390,17 +900,61 @@ But first we have a few improvements to make to our setup.py so that
...
+'/Users/tpratchett/anaconda3/lib/python3.6/site-packages']
+```
+
+The first will be your current directory - a Python environment
+always looks there first -
+and the rest are system paths for
+the Python installation you are using.
+
+To install a package simply means to copy the contents of the
+package in some form into one of the directories that Python will search in.
```
lwjohnst86:
```suggestion
Installing a package means that the contents of the package are copied
into one of the directories that Python will search in.
```

py-rse/package-py.Rmd
```diff
@@ -1421,3 +975,166 @@ if (knitr::is_latex_output()) {
...
+Since any code that does an `import zipf` will read `__init__.py`, it
+can be convenient to put import statments in `__init__.py` itself, to
+ensure that some key definitions are readily available when the package
+is imported.
+
+We'll try that here:
+
+1. Add the lines `from .generate import *` and `from .check import *` to
+`zipfpy/__init__.py`.
+2. In the test suite, you can now just call `import zipfpy` and refer to,
```
lwjohnst86:
```suggestion
2. In the test suite, you can now call `import zipfpy` and refer to,
```

lwjohnst86: "just" :wink:

py-rse/package-py.Rmd
```diff
@@ -1421,3 +975,166 @@ if (knitr::is_latex_output()) {
...
+
+### Exercise: What is `__name__` in different contexts?
+
+1. Start an interactive Python interpreter, and run `print(__name__)`.
+What do you get?
+2. Add the statement `print(__name__)` to the end of the file `zipf.py`
+and from the Python interpreter, run `import zipf`.  What is the name
+printed?
+
+### Exercise: Turn `use.py` into a unit test
```
lwjohnst86: The next chapter will be unit testing, not before. So no reference to unit testing should be included.

py-rse/package-py.Rmd
lwjohnst86: Also, this could be moved to the correctness chapter, since this comes after the packaging chapter.

py-rse/package-py.Rmd
```diff
@@ -20,445 +20,110 @@ This lesson shows you how to use Python's tools to create and share libraries of
...
+It might initially be confusing
+that the top level directory has the same name as the package directory containing the `.py`-files.
+You could change the name of the top level folder to whatever you want,
+but it is often convenient that this is the same as the package directory.
+If you change the name of the package directory,
+this will change the name of your package.
+You could also place your package_directory inside a subdirectory,
+such as `src` or `py`,
+but this would require some changes to the setup instructions later on,
+so we will stick to the conventional folder hierarchy above.
```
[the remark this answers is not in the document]
joelostblom: Ok, I will wait until we have finished the discussion of which is preferred of `pkg_name/pkg_name` or `pkg_name/src/pkg_name`

py-rse/package-py.Rmd
```diff
@@ -20,445 +20,110 @@ This lesson shows you how to use Python's tools to create and share libraries of
...
+
+The folder contains a Python file with a special name: `__init__.py`.
+This was a requirement before Python 3.3,
+and is now included as a convention
+or to execute code upon package import.
```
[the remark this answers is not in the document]
joelostblom: I think the original intent with this sentence is that some people just add an empty `__init__.py` because that is how it used to be and some people add an `__init__.py` specifically to execute the code it contains on import.

py-rse/package-py.Rmd
```diff
@@ -20,445 +20,110 @@ This lesson shows you how to use Python's tools to create and share libraries of
...
+from the local copy source code using standard tools!
+To install this package,
+we could `cd` to the same directory that contains `setup.py`
+and type `pip install .`
+(where `.` references the current directory).
+However,
+it is considered good practice to create a separate, virtual Python environment
+when installing our own packages,
+since this will allow us to test how your package installs in a clean Python environment,
+rather that the current state of your primary Python environment.
```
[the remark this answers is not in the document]
joelostblom: Yes I think CI should also cover this. Personally, I would create a clean env for my package anyways when I am developing it, but we could reword (or I could be enlighten on why I don't need to do this).

py-rse/package-py.Rmd
joelostblom: Opened #341

lwjohnst86:
merging in :)

joelostblom:
@DamienIrving @gvwilson Luke and I had a misunderstanding regarding the merge, so I have opened #354 as a continuation of this PR, but just to address the points we have been discussing here:

1. Done
2. Done
3. Done
4. Done
5. Created issue #339
6. I think we should start with the correct dir structure. Include a simple structure and a one line readme before developing any code so that students have an overview on where things goes.

joelostblom:
also on 5 above, shebangs were not needed with the entry points syntax

Issue #340, opened 2020-03-11
Discussion: Persona last names, or other ideas for people in Git chapters


cwickham:
The git lessons used the member of the ENIAC project as users:
* [Frances Bilas in `git-cmdline.Rmd`](https://github.com/merely-useful/merely-useful.github.io/blob/aaef6cf97996e00f6d03282ed2320edc1a6a4f98/py-rse/git-cmdline.Rmd#L22-L25)
* [Frances Bilas and Jean Jennings in `git-advanced.Rmd`](https://github.com/merely-useful/merely-useful.github.io/blob/aaef6cf97996e00f6d03282ed2320edc1a6a4f98/py-rse/git-advanced.Rmd#L653-L657)

I've used one of our [learner personas](https://merely-useful.github.io/py-rse/index.html#rse-intro-personas), Amira, in my rework of `git-cmdline.Rmd`, but [I need a lastname](https://github.com/merely-useful/merely-useful.github.io/pull/306#discussion_r386080200).  **How do people feel about "Amira Khan"?**

@SaraMati will need a second user for `git-advanced.Rmd`.  Should it be another one of our personas, Sami?  **What should their last name be?**

joelostblom:
Khan sounds good. Following the same royal reasoning we could have "Sami Wang", although I personally like the sound of "Sami Weaver" which was just a name that popped up when I looked around. Or something Finnish...

DamienIrving:
I guess we want to pick surnames that typically go with those first names? If so, here's what google tells me about those first names:

## Amira
"Amira is an Arabic female given name, meaning "princess", and a Hebrew female given name, meaning 'treetop' or 'saying'."

Khan is a very common surname in the Arab world, so Amira Khan sounds good.

## Sami
"Sami can be an Arabic name meaning elevated or sublime, or a Turkish name, a Finnish male name derived from Samuel, or an American name abbreviated from Samantha or Samuel."

The pronouns used for Sami in the personas are they/them, so we can interpret it as a male or female given name. If we went with Joel's suggestion of a Finnish surname we could go with Sami Nieminen or Sami Heikkinen or something like that?

cwickham:
To keep a complete record, from @gvwilson in Slack:
> how about "Sorva" as a last name? "Sami Sorma" would be Finnish...

DamienIrving:
We also need a last name for Jun...

gvwilson:
Jun Hsu ?

[...]

PR #349, opened 2020-04-02
Updating advanced git chapter with Zipf's content

DamienIrving:


DamienIrving:
This is ready to review.

A few things to note:  
- Once the content of the command line git and advanced git chapters are settled, we'll need to create a `zipfs-law` git repo and go through each commit one-by-one so we can get a consistent commit history. We need that for generating the `git log` output (down to the precise SHA-1 hash identifier for each commit) and we'll also need to link that repo to https://github.com/amira-khan/zipfs-law (I've created a dummy GitHub account for Amira) so that we can get all the screenshots for chapters.
- We need to settle on Sami's surname (#340) so I can create a dummy GitHub account for them
- The very last section of this chapter is on tags and semantic versioning. I'm thinking that would be better placed in the packaging section? @joelostblom #354

joelostblom:
@DamienIrving Regarding semantic versioning. I think it could be introduced in either place. It makes sense to have it here because we are introducing tags and then talking about how to name them feels natural to me (there are of course other viable names depending the project nature, but for packages it makes sense to primarily talk about semantic versioning I think). 

Btw, git introduced and recommend to use `restore` and `switch` over some usage of `checkout` [somewhat recently](https://medium.com/blue-harvest-tech-blog/git-2-23-0-forget-about-checkout-and-switch-to-restore-ac2682b737b3). I think we don't need more content to add now, but I just wanted to note this, especially since `git status` directly mentions to use `restore`, so that we are aware and it cold be a future update.

DamienIrving:
Thanks, @SaraMati. I think going through the commits one-by-one in Amira's GitHub repo will have to be done as a very final step in producing this volume, because if we alter any of the code changes that are committed (or the order in which they are committed) during the editing process then we'd have to do the whole thing over again. I've created an issue so we don't forget: #359 

@joelostblom Since you don't mind either way, I've moved semantic versioning to the packaging chapter. I've also removed the content on tags - there's already so much stuff in this chapter and I don't think it's critical people know about them. Thanks also for the heads up about `git swtich` and `git restore`. I agree that we don't need to add them yet (sounds like they are a test as opposed to permanent feature at this stage) but down the track we might.

py-rse/git-advanced.Rmd
```diff
@@ -116,28 +181,253 @@ We can (and should) choose the name to signal the purpose of the branch,
...
+        Scatter plot to which the power curve will be added.
+    """
+    xvals = np.arange(curve_xmin, curve_xmax)
+    yvals = max_rank * (xvals**(-beta + 1))
+    ax.loglog(xvals, yvals, color='grey')
+```
+
+where the maximum word frequency rank corresponds to \(a\)
+in the general form of a power law and \(-\beta + 1\)
+corresponds to \(b\).
```
cwickham, line 299:
```suggestion
where the maximum word frequency rank corresponds to intercept
in our linear form, and \(-\beta + 1\) the slope.
```

py-rse/git-advanced.Rmd
```diff
@@ -116,28 +181,253 @@ We can (and should) choose the name to signal the purpose of the branch,
...
+        Maximum x-bound for fitted curve
+    max_rank : int
+        Maximum word frequency rank.
+    beta : float
+        Estimated beta parameter for the power law.
+    ax : matplotlib axes
+        Scatter plot to which the power curve will be added.
+    """
+    xvals = np.arange(curve_xmin, curve_xmax)
+    yvals = max_rank * (xvals**(-beta + 1))
```
cwickham, line 293: Maybe we should rewrite this to use `alpha` not `beta`, so it links better with the previous section.

DamienIrving: That's probably a good idea. I guess we can wait until you resolve #367 to rewrite, since the rewrite also needs to acknowledge the variable C which way say we'll refer back to but don't as yet.

DamienIrving:
Thanks, @cwickham! I accepted all your comments that don't relate to #367, so I guess this one is ready to merge.

PR #350, opened 2020-04-03
Getting a clean build:

gvwilson:
1.  Use 'book' as documentclass instead of 'Nemilov' to circumvent problems with Unicode.
2.  Add a bunch of shrapnel to `.gitignore`.

### DIFF
--- a/_py-rse.yml
+++ b/_py-rse.yml
@@ -32,6 +32,7 @@ rmd_files:
   - CONDUCT.md
   - CONTRIBUTING.md
   - gloss.md
+  - py-rse/install.Rmd
   - py-rse/objectives.Rmd
   - py-rse/keypoints.Rmd
   - py-rse/solutions.Rmd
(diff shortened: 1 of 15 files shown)


gvwilson:
I have added `preamble.tex` to force the page style to be plain, which removes the warnings about ALL CAPS section names in the page header; I had to update `_output.yml` as well to include this file, and `Makefile` to make it a dependency.

py-rse/install.Rmd
```diff
@@ -0,0 +1,3 @@
+# Setting Up {#py-rse-install}
+
+FIXME: write installation instructions.
```
lwjohnst86, line 3: Sorry, I'm not sure what this means. To install the book and dependencies? For another chapter?

_py-rse.yml
```diff
@@ -32,6 +32,7 @@ rmd_files:
   - CONDUCT.md
   - CONTRIBUTING.md
   - gloss.md
+  - py-rse/install.Rmd
```
lwjohnst86, line 35: What is this for?

PR #351, opened 2020-04-04
Revisions to Python RSE style chapter.

### LINKED ISSUE #149: Revise code style and review
Revise `style.Rmd`:

* Exercises
* R components?
* Add section on review

### LINKED ISSUE #190: Raw material for code review
- https://the-turing-way.netlify.com/reviewing/reviewing.html
- https://github.com/alan-turing-institute/the-turing-way/issues/656

### LINKED ISSUE #271: add short section to rse-style.Rmd showing how to do code review on GitHub
Unless it would be more appropriate to show it in the novice volumes.

gvwilson:
- Closes #149 
- Closes #190 
- Closes #271


py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
-:   "Did you maybe think about *testing* this garbage?"
-    is a Code of Conduct violation in any well-run project.
+Python automatically interprets `a+b<c+d` as `(a+b)<(c+d)`,
+but that's a lot of punctuation crowded together;
+using spaces around comparison operators makes it easier to see
+what's being compared to what.
+However,
+you should use your own judgment for spacing around arithmetic operators like `+` and `/`.
+For example,
+`a+b+c` is perfectly readable,
```
joelostblom, line 84: Perhaps mention that spacing can effectively be used to indicate priority without having to clutter with parenthesis? As in `a*b + c`.

py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
-:   Nobody enjoys fixing bugs and style violations.
-    Asking them to add a few features while they're at it is rude.
+**Put two blank links between each function definition.**
 
-Don't let people break these rules just because they're frequent contributors or in positions of power.
-:   The culture of any organization is shaped by the worst behavior it is willing to tolerate @Grue2015.
-    If you let people be rude to one another,
-    *that* is your culture.
+This helps the eye see where one ends and the next begins,
+though the fact that functions always start in the first column helps as well.
```
joelostblom, line 101: In the spacing section, I would also mention spaces around variable assignment, but not parameter assignment. Mixing these up is a common novice mistake in my experience.

py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
+which are outside the scope of this lesson.
 
-FIXME: move this to the novice material.
+**Avoid abbreviations in function and variable names.**
 
-### Always indent code blocks using 4 spaces, and use spaces instead of tabs.
+They can be ambiguous,
+and can be be hard for non-native speakers to understand.
+This rule doesn't necessarily you will have to do more typing:
+a good programming editor will [auto-complete][auto-completion] names for you.
```
joelostblom, line 128: Hmmm, could we also have a note that this is not an OK signal to use giant variable names, that is really hard to read as well.

py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
+provided the loop is only a few lines long @Beni2017.
+Anything that is used at a greater distance
+or whose purpose isn't immediately clear
+(such as a function) should have a longer name.
 
-### Always use spaces around comparisons like `>` and `<=`.
+**Keep functions short.**
 
-Use your own judgment for spacing around common arithmetic operators like `+` and `/`.
+No function should be more than a page long
```
joelostblom, line 140: What's a page here? A screen page?

py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
+others of which are utilities that may be called many times from many different places.
+We put all of the single-use functions in the first half of the file
+in the order in which they are likely to be called,
+and then put all of the multi-use utility functions in the bottom of the file in alphabetical order.
+If any of those utility functions are used by other scripts or programs,
+they should go in a file of their own.
+In fact,
+this is a good practice even if they're only used by one program,
+since it signals even more clearly which functions are specific to this program
+and which are likely to be reused elsewhere.
```
joelostblom, line 255: Should we include anything about breaking into multiple files?

py-rse/style.Rmd
```diff
@@ -338,65 +386,115 @@ if __name__ == '__main__':
     print('number', n)
 ```
 
-## How should I collaborate with people when doing code reviews? {#py-rse-style-how}
+## How should I review someone else's code? {#py-rse-style-review}
 
-FIXME: how to review on GitHub.
+The biggest benefit of having a second person work on a programming project
+is therefore not getting twice as much code written,
```
joelostblom, line 392:
```suggestion
is not getting twice as much code written,
```

py-rse/style.Rmd
```diff
@@ -338,65 +386,115 @@ if __name__ == '__main__':
...
+but having code reviewed.
+Study after study over more than 40 years has shown that code review is the most effective way to find bugs in software
+@Faga1976,@Faga1986,@Cohe2010,@Bacc2013.
+Despite this,
+it still isn't common in research software development,
+in part because it isn't part of the culture @Sega2005,
+but also because code review is mostly useful
+when the reviewers understand the problem domain well enough to comment on algorithms and design choices
+rather than indentation and variable naming,
```
joelostblom, line 401: :100:

py-rse/style.Rmd
```diff
@@ -338,65 +386,115 @@ if __name__ == '__main__':
...
-4.  Write output.
+Work in small increments.
+:   As @Cohe2010 and others have found,
+    code review is most effective when done in short bursts.
+    That means that change requests should also be short:
+    anything that's more than a couple of screens long
+    should be broken into smaller pieces.
 
-or:
+Look for algorithmic problems first.
```
joelostblom, line 426: Somewhere it would be good to add a note that following coding coding style helps you reviewers focus on algorithmic problems.

py-rse/style.Rmd
```diff
@@ -338,65 +386,115 @@ if __name__ == '__main__':
...
-In fact,
-this is a good practice even if they're only used by one program,
-since it signals even more clearly which functions are in the "structural" layer
-and which are in the "utility" layer.
+Don't let people break these rules just because they're frequent contributors or in positions of power.
+:   The culture of any organization is shaped by the worst behavior it is willing to tolerate @Grue2015.
+    If you let people be rude to one another,
+    *that* is your culture.
+
+Be specific in replies to reviewers.
```
joelostblom, line 485: Do you think it would be helpful to add an additional heading for how to receive a review? It would include the two points below and maybe also something along the lines of:

> Give people the benefit of the doubt, someone might have left comments that seem short but people (especially those volunteering in open source projects) often have little time to write elaborate replies to everyone's questions and suggestions. Asking for clarification is more productive than becoming defensive, often no harm is meant.

Rephrase to sound better.

code-review/stop-words.py
```diff
@@ -0,0 +1,23 @@
...
+stops = ['a', 'A', 'the', 'The', 'and']
+
+
+def count(ln):
+    n = 0
+    for i in range(len(ln)):
+        line = ln[i]
+        stuff = line.split()
+        for word in stuff:
+            # print(word)
```
gvwilson, line 13: Remove or turn into logging statement.

code-review/stop-words.py
```diff
@@ -0,0 +1,23 @@
+import sys
+
+
+stops = ['a', 'A', 'the', 'The', 'and']
+
+
+def count(ln):
+    n = 0
+    for i in range(len(ln)):
```
gvwilson, line 9: Should iterate over contents of list rather than range of indices.

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
-named_parameters = {'high': 'upper bound'}
-trim_value(*parameters, **named_parameters)
+scale = 0.8 if configuration['threshold'] > UPPER_BOUND else 1.0
 ```
 
-```text
-some matrix with lower bound and upper bound
-```
+However,
+this puts the default last instead of first,
```
joelostblom, line 624: But you could just use `if not` to put the default first, so is this really a drawback significant enough to mention?

---

Btw, sometimes I wish Python had `unless` and `then` instead of `if not` and `else`, it would read so nicely =)

```ruby
scale = 0.8 unless configuration['threshold'] > UPPER_BOUND then 1.0
```

See how nice it looks! (turns out ruby has this, that's why the syntax highlighting is correct above)

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
+scale = 1.0
+if configuration['threshold'] > UPPER_BOUND: scale = 0.8
+```
 
-We can use the inverse of `*args` and `**kwargs` to match a list of values to arguments.
-In this case,
-we put the `*` in front of a list and `**` in front of a dictionary when *calling* the function,
-rather than in front of the parameter when *defining* it:
+Some programmers take this even further
+and use a [conditional expression][conditional-expression]:
```
joelostblom, line 617:
```suggestion
and use a [conditional expression][conditional-expression] (also referred to as a "ternary operator"):
```

To help searching for more info?

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
 
-The comparison to `True` is unnecessary because `a > b` is a Boolean value
-that is itself either `True` or `False`.
-Like any other value,
-Booleans can be assigned to variables,
-and those variables can then be used directly in tests:
+The expressions used to update `num_vowels` and `num_consonants` make use of the fact that
+`in` produces either `True` or `False`,
+which the function `int` converts to either 1 or 0.
+We will explore ways of making this code more readable in the exercises.
```
joelostblom, line 778: Is it worth mentioning somewhere that Python does not have a "case" operator, but that lookup tables or if/else are used for that instead.

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
+    n = count(sys.stdin)
+    print('number', n)
 ```
 
-is much easier to read than the equivalent expression:
+Reading code in order to find bugs and suggest improvements like these
+is called [code review][code-review].
+Study after study over more than 40 years has shown that code review is the most effective way to find bugs in software
+@Faga1976,@Faga1986,@Cohe2010,@Bacc2013.
+It is also a great way to transfer knowledge between programmers:
```
joelostblom, line 872: Yes, this is so important and effective!

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
+
+Despite this,
+code review still isn't common in research software development.
+This is partly a chicken-and-egg problem---people don't do it
+because other people don't do it @Sega2005.
+It is also more difficult to do in specialized scientific fields:
+in order for review to be useful,
+reviewers need to understand the problem domain well enough to comment on algorithms and design choices
+rather than indentation and variable naming,
+and the number of people who can do that for a research project is often very small @Petr2014.
```
joelostblom, line 884: Also, in fields where software development is used merely as a tool and not an end goal itself, it is rarely seen as important enough by supervisors, so they will likely not ask any of their grad students to spend time on it. Or they only have one grad student that can code in the first place and see not point is getting more.

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
+def collect(new_value, accumulator=set()):
+    accumulator.add(new_value)
+    return accumulator
 ```
 
-In simple cases,
-people will sometimes put the test and assignment on a single line:
+it calls `set()` to create a new empty set,
+and then uses that set as the default value for `accumulator` every time the function is called.
+It does *not* call `set()` once for each call,
```
joelostblom, line 1070: Mention that this is called "mutable defaults" and link either the docs (https://docs.python.org/3/library/dataclasses.html#mutable-default-values was the best I could find) or something clearer like http://effbot.org/zone/default-values.htm

py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
+which again was more than most people felt necessary.
+Today,
+almost all editors will auto-indent or auto-complete when the tab key is pressed,
+but the legacy of those times lives on.
+
+**Do not put spaces inside parentheses.**
+
+Don't write <code>(&nbsp;1+2&nbsp;)</code>.
+This applies to function calls as well:
+write `max(a, b)` rather than <code>max(&nbsp;a,&nbsp;b&nbsp;)</code>.
```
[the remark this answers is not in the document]
lwjohnst86: We don't. Markdown code will render the extra spaces.

py-rse/style.Rmd
```diff
@@ -59,165 +48,228 @@ Look for algorithmic problems first.
...
-and PEP-8 has become the most widely used standard for Python coding.
-Some of its rules are listed below,
-along with others borrowed from "[Code Smells and Feels][code-smells-and-feels]":
+**Use `lower_case_with_underscores` for the names of functions and variables.**
 
-FIXME: add before-and-after examples for each of these.
+Research on naming conventions has produced mixed results @Bink2012,@Scha2018,
+but Python has (mostly) settled on underscored names for most things.
+This naming convention is called [snake case][snake-case] or [pothole case][pothole-case];
+you should only use [`CamelCase`][camel-case] for classes,
```
DamienIrving, line 123: In #354  "pothole case" is referred to as "kebab case"

py-rse/style.Rmd
```diff
@@ -59,165 +48,228 @@ Look for algorithmic problems first.
...
--   The file's documentation string (Chapter \@ref(py-rse-package-py-document)).
+-   The file's documentation string (Section \@ref(py-rse-package-py-document)).
 -   All of the `import` statements, one per line.
--   Constant definitions.
+-   Global variable definitions
+    (especially things that would be constants in languages that support them).
 -   Function definitions.
 -   If the file can be run as a program,
     the `if __name__ == '__main__'` statement discussed in
     Section \@ref(py-rse-package-py-import).
```
DamienIrving, line 209: It's section `py-rse-py-scripting-main`

py-rse/style.Rmd
```diff
@@ -59,165 +48,228 @@ Look for algorithmic problems first.
...
+but programmers disagree (strongly) on whether high-level functions should come first or last,
+i.e.,
+whether `main` should be the first function in the file or the last one.
 
-If you need to do something in some runs of the program and not do it in others,
-use an `if` statement to enable or disable that block of code.
-It's much more reliable—you're far less likely to accidentally comment out one too many lines—and
-you may find that you want to leave those conditional sections in the finished program
-for logging purposes (Section \@ref(py-rse-errors-logging)).
+We prefer the former approach,
```
DamienIrving, line 215: As it stands all our scripts (i.e. `countwords.py`, `collate.py` and `plotcounts.py` have `main` at the bottom of the file, so everything (the scripts and all the references to their contents them in the early chapters) will need to be updated accordingly. 

I'm thinking that could be a separate issue / PR rather than including it in this PR.

py-rse/style.Rmd
DamienIrving: We might also want to discuss how the code we've developed in the preceding chapters (i.e. 'countwords.py', 'collate.py' and 'plotcounts.py', plus the soon to be renamed `mymodule.py`) applies these patterns.

py-rse/style.Rmd
```diff
@@ -59,165 +48,228 @@ Look for algorithmic problems first.
...
-Python has a tool that used to be called `pep8`
-and is now called `pycodestyle`
-that will do this.
-For example,
-this program is supposed to count the number of stop words in a document:
+
+Python's linter used to be called `pep8` and is now called `pycodestyle`.
+To see how it works,
+let's look at this program,
+which is supposed to count the number of [stop words][stop-word] in a document:
```
DamienIrving, line 272: Instead of defining a new program, perhaps we should run the linter over one of the shorter scripts we've developed (`countwords.py` or `collate.py`)?

DamienIrving: The downside of course is that we'd need to write sub-optimal code earlier on in order for the linter to pick something up...

py-rse/style.Rmd
```diff
@@ -4,52 +4,41 @@
...
-it still isn't common in research software development,
-in part because it isn't part of the culture @Sega2005,
-but also because code review is mostly useful
-when the reviewers understand the problem domain well enough to comment on algorithms and design choices
-rather than indentation and variable naming,
-and the number of people who can do that for a research project is typically very small—sometimes
-as small as one @Petr2014.
+but people do if they are to create, debug, and extend them.
+This lesson therefore presents guidelines for good Python programming style
+and introduces some language features that can make programs more flexible and more readable.
```
DamienIrving, line 14: Perhaps this introductory paragraph should set the scene by saying something like the following? 

"In the preceding chapters we've written a bunch of code to analyze the word counts of classic english novels. We're going to take a step back in this chapter and discuss some of the style choices that were made in developing that code and why."

py-rse/style.Rmd
```diff
@@ -4,207 +4,255 @@
...
+(which we find perfectly readable, but some people find too crowded)
+and 8
+(which most people agree uses up too much horizontal space).
+As for the use of spaces rather than tabs,
+the original reason was that the most common interpretation of tabs
+by the editors of the 1980s was 8 spaces,
+which again was more than most people felt necessary.
+Today,
+almost all editors will auto-indent or auto-complete when the tab key is pressed,
+but the legacy of those times lives on.
```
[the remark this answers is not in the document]
gvwilson: parenthetical comment added.

py-rse/style.Rmd
gvwilson: I think this is bad style: `a*b + c*d` is OK, but what how do you add spaces to `a**2*b+c**2*d`? Safer to tell people to parenthesize...

py-rse/style.Rmd
gvwilson: Done - cross-references added.

py-rse/style.Rmd
gvwilson: Is there anything in PEP-8 about overly-long names?

py-rse/style.Rmd
gvwilson: printed page - rule updated

py-rse/style.Rmd
gvwilson: How to break code between files feels like it belongs somewhere near here, but I'm worried that this chapter is already a long catalog that readers won't be able to absorb. Could it go with packaging?

py-rse/style.Rmd
gvwilson: done.

py-rse/style.Rmd
gvwilson: Done.

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
+# ...before...
+def rescale_by_average(values, factors, weights):
+    a = 0.0
+    for (f, w) in zip(factors, weights):
+        a += f * w
+    if a == 0.0:
+        return
+    a /= len(f)
+    if not values:
+        return
```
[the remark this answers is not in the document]
gvwilson: added.

py-rse/style.Rmd
```diff
@@ -287,755 +349,875 @@ n = count(lines)
...
+and use a [conditional expression][conditional-expression]:
 
 ```python
-def trim_value(data, low, high):
-    print(data, "with", low, "and", high)
-
-parameters = ['some matrix', 'lower bound']
-named_parameters = {'high': 'upper bound'}
-trim_value(*parameters, **named_parameters)
+scale = 0.8 if configuration['threshold'] > UPPER_BOUND else 1.0
```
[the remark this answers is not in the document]
gvwilson: Not sure what change you're asking for?

py-rse/style.Rmd
gvwilson: I think most of our readers will never have run into `case`.

py-rse/style.Rmd
gvwilson: note added.

py-rse/style.Rmd
gvwilson: Does that jargon add value for our readers?

py-rse/style.Rmd
```diff
@@ -4,220 +4,272 @@
...
+```
 
 Look for algorithmic problems first.
 :   Code review isn't just (or even primarily) about style:
     its real purpose is to find bugs before they can affect anyone.
     The first pass over any change should therefore look for algorithmic problems.
     Are the calculations right?
     Are any rare cases going to be missed?
-    Are errors being caught and handled Chapter \@ref(logging)?
+    Are errors being caught and handled (Section \@ref(py-rse-errors-logging))?
```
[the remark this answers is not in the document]
gvwilson: done.

py-rse/style.Rmd
```diff
@@ -4,220 +4,272 @@
...
+the original reason was that the most common interpretation of tabs
+by the editors of the 1980s was 8 spaces,
+which again was more than most people felt necessary.
+Today,
+almost all editors will auto-indent or auto-complete when the tab key is pressed,
+but the legacy of those times lives on.
+
+**Do not put spaces inside parentheses.**
+
+Don't write <code>(&nbsp;1+2&nbsp;)</code>.
```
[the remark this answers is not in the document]
gvwilson: converted to non-breaking space characters.

py-rse/style.Rmd
```diff
@@ -4,220 +4,272 @@
...
-    comment on the first two or three and simply put a check beside the others—the reader
-    won't need the comment repeated.
+```python
+substrate[i, j]+overlay[i, j]
+```
 
-Don't ask people to do extra work.
-:   Nobody enjoys fixing bugs and style violations.
-    Asking them to add a few features while they're at it is rude.
+**Put two blank links between each function definition.**
```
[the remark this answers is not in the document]
gvwilson: I took it out because it added a lot of space on the page without the readability difference being clear - I think it only pays off once you have many functions.

py-rse/style.Rmd
```diff
@@ -4,220 +4,272 @@
...
+you should only use [`CamelCase`][camel-case] for classes,
+which are outside the scope of this lesson.
 
-FIXME: move this to the novice material.
+**Avoid abbreviations in function and variable names.**
 
-### Always indent code blocks using 4 spaces, and use spaces instead of tabs.
+They can be ambiguous,
+and can be be hard for non-native speakers to understand.
+This rule doesn't necessarily you will have to do more typing:
```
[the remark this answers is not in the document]
gvwilson: good catch.

py-rse/style.Rmd
```diff
@@ -4,220 +4,272 @@
...
+in the order in which they are likely to be called,
+and then put all of the multi-use utility functions in the bottom of the file in alphabetical order.
+If any of those utility functions are used by other scripts or programs,
+they should go in a file of their own.
+In fact,
+this is a good practice even if they're only used by one program,
+since it signals even more clearly which functions are specific to this program
+and which are likely to be reused elsewhere.
 
 ## How can I check that code follows style guidelines? {#py-rse-style-check}
```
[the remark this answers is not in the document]
gvwilson: every Python IDE I've used will run PEP-8 - can someone who's using an IDE address this one?

py-rse/style.Rmd
gvwilson: thanks - fixed.

py-rse/style.Rmd
gvwilson: I will adjust the comment instead.

py-rse/style.Rmd
gvwilson: That was my thinking...

py-rse/style.Rmd
gvwilson: added

py-rse/style.Rmd
joelostblom: I think the added value is that the name and link are useful for leaning more, and that the mention of the term here where we clearly explain it makes it less intimidating to come across it later (where it might not have a good explanation).

py-rse/style.Rmd
joelostblom: Same comment here as below on jargon.

py-rse/style.Rmd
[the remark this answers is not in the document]
joelostblom: This is introduced following the "default and override" paragraph, but I think it is more similar to if/else, so might be more logical to but after that paragraph?

py-rse/style.Rmd
joelostblom: I was thinking something short, a sentence or two, but I am also fine with leaving as is since we do mention when to break into files. If we were to put it elsewhere, I think it would go best with where we create the second script together with an explanation of why we didn't just continue on the first script.

py-rse/style.Rmd
joelostblom: Nope, my comment was about readability in general. To be clear, I am not saying that abbreviations should be used when a variable name exceeds a certain length, but that if you can think of a shorter name that is as descriptive, then use that instead. It should be clear and concise, and use commonly accepted abbreviations (e.g. "id" instead of "identification_number").

py-rse/style.Rmd
joelostblom: I meant spaces as a complement to parenthesis. In simple examples, only spaces offers more readability than parenthesis in my opinion, and in more complex cases both can be used so it doesn't get cluttered with too many parentheses. If I wanted to clarify the expression you gave, I would write it as 

```
(a**2)*b + (c**2)*d
```

which I think is clearer than both 
```
(a**2)*b+(c**2)*d
```

and

```
((a**2)*b)+((c**2)*d)
```

gvwilson: PEP-8 doesn't appear to mandate anything on this, and since we differ on what to recommend, I feel we should leave this as is.

py-rse/style.Rmd
gvwilson: Again, PEP-8 doesn't make a recommendation, and I would personally prefer `identification_number` over `id` for a variable with wide scope...

py-rse/style.Rmd
gvwilson: Can you point me at where we mention breaking code into files? And which script do you mean by "second script"?

py-rse/style.Rmd
gvwilson: comment added.

py-rse/style.Rmd
gvwilson: I'd never actually run into the phrase "mutable defaults" before you mentioned it - how do others feel?

py-rse/style.Rmd
joelostblom: Above at line 250-251:

> If any of those utility functions are used by other scripts or programs,
> they should go in a file of their own.

By second script I was referring to when we in the Python scripting chapter decide to create a new file for some of the functionality instead of keep adding to the first file we created. In the end we have four files (countwords, collate, plotcounts, mymodule), but we could have had one really long file.

py-rse/style.Rmd
joelostblom: To me, this paragraph is about general best practices in coding, not limited to those mentioned in PEP 8 (note that PEP 8  does not recommend against using abbreviations in names either).

I agree that verbosity depends on scope, but there is also meaningless verbosity that doesn't convey more information, which is where I think `id` falls (another example would be expanding `usa` to `united_states_of_america`, and [this HN comment](https://news.ycombinator.com/item?id=12311031) illustrates and example not related to abbreviations). This [SO answer](https://stackoverflow.com/questions/2230871/when-is-a-java-method-name-too-long/2230893#2230893) sums up how I think of this (without saying it is more right or wrong because it is on SO):

> A name in Java, or any other language, is too long when a shorter name exists that equally conveys the behavior of the method.

I think the most important with this recommendation is to discourage authors from creating their own abbreviations, which might be clear in the moment they write them, but not after that or for anyone else reading the code.

py-rse/style.Rmd
joelostblom: This is actually in PEP 8, third bullet under [Other Recommendations](https://www.python.org/dev/peps/pep-0008/#other-recommendations):

> If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator:
>
> ```python
> # Correct:
> i = i + 1
> submitted += 1
> x = x*2 - 1
> hypot2 = x*x + y*y
> c = (a+b) * (a-b)
> ```
> 
> ```python
> # Wrong:
> i=i+1
> submitted +=1
> x = x * 2 - 1
> hypot2 = x * x + y * y
> c = (a + b) * (a - b)
> ```

A gotcha might worth mentioning is that while whitespace conveys priority to the reader, it does not convey it to the Python interpreter so it can't be treated the same as surrounding parentheses.

py-rse/style.Rmd
gvwilson: Gotcha - added

PR #356, opened 2020-04-13
todo package for all those TODO items

tallguyjenks:
leaves a really obvious margin note like MS Word's track changes with notes so you can see items you need to resolve. made with `\todo{note here}`


lwjohnst86:
Hi @tallguyjenks thanks for the PR! We actually use HTML for the most part for drafting and revising, so unless this works from `md` to `html` conversion, it wouldn't fit our purposes.

Issue #344, opened 2020-03-16
Check when and where we orignally create zipfs-law/ and refer to it consistently


joelostblom:
Now it is a mix of `zipfs-law/` and `~/zipfs-law/` in different chapters.

k8hertweck:
Resolved by #363

Issue #310, opened 2020-01-17
Discussion: Add the output from `tree` in the beginning of each/some RSE chapters?


joelostblom:
To orient students about what the current file structure should looks like. I think this is important for the chapters that builds on files being created in the previous chapters.

lwjohnst86:
Before we have this, we would need to have a solid "path" between chapters of what is being created etc. Maybe make individual folders based on what the package should look like?

DamienIrving:
As well as the current state of the project directory structure, some chapters may also need to provide the current state of scripts that are under development (e.g. the `plotcounts.py` script).

PR #363, opened 2020-04-23
py-rse standardization of files across chapters for zipfs-law, including project structure chapter

k8hertweck:
In working through the project structure chapter, I realized I needed to work through from the beginning to ensure everything has been standardized as much as possible. 

This PR assumes that readers will download and start from a directory called `zipfs-law` at that includes:

```
└── data
    ├── README.md
    ├── dracula.txt
    ├── frankenstein.txt
    ├── jane_eyre.txt
    ├── moby_dick.txt
    ├── sense_and_sensibility.txt
    ├── sherlock_holmes.txt
    └── time_machine.txt
```

Changes here include:
- Noting when there are files/directories created (such as in the bash chapters) that aren't actually used later, and ignoring them in the `tree` outputs at the beginning of other chapters.
- adding FIXME in places where it's not entirely clear what a reader should do to proceed (e.g., is this an example for illustration, or something required for the zipfs-law project?)

I'm happy to resolve additional concerns (such as the section in the introduction describing where the files come from).


k8hertweck:
This PR attempts to reconcile the files included in all chapters, and includes an output from `tree` for most chapters (resolving #310 ). Remaining issues about structure and content include:
- I didn't include `tree` output for the final chapter on publishing; this chapter has more overlap than I anticipated with the project structure chapter and should be reconciled
- some chapters still lack integration with the main narrative. Some, like style, may not need much reconciliation. Others, particularly correctness, could use quite a lot of work.
- is the project named `zipfs-law` or `zipfs_law`? The former is used more frequently, but the latter is required by packaging (dashes aren't allowed).
- the `zipfs-law/` directory in the repo needs to be updated to match naming (above), as well as extraneous files (makefiles) removed
- images, references, and glossary could all use work
- many remaining exercises are incomplete (and some exercises are present in bash chapters, but location of data/files to answer them isn't specified)

Along the way, I did some minor copy editing and noted the following stylistic things that should be cleaned up (but I didn't want to make this PR more confusing):
- standardize Zipf's Law (spelling of Zipf's, capitalization, apostrophe, etc)
- use code formatting for paths mentioned in text, including / after directories mentioned outside code chunks
- pop-out boxes should include titles in bold (in title case)

py-rse/bash-basics.Rmd
```diff
@@ -302,7 +302,7 @@ $ ls -F zipfs-law
 ```
 
 ```text
-bin/  data/  Makefile  makefiles/
+data/
```
gvwilson, line 305: Does it really only contain the `data` folder at this point? (Looking down further: yup, it appears so.)

py-rse/bash-basics.Rmd
```diff
@@ -440,11 +440,14 @@ but we'll see some uses for it soon.
...
-> In these cases,
-> it's easier to combine them together behind a single dash:
+> In most command line tools,
+> multiple options can be combined with a single `-`
+> and no spaces between the options:
+>
 > ```shell
 > $ ls -Fa
 > ```
+>
```
gvwilson, line 450: I would add a line saying that this isn't recommend practice because some commands take long-name options and it's very easy to mistake `-no` (meaning `-n -o`) with `--no` (meaning "answer 'no' to all questions").

py-rse/bash-basics.Rmd
```diff
@@ -1324,12 +1328,17 @@ $ wc -l *.txt | sort -n | head -n 3
   13028 sense_and_sensibility.txt
 ```
 
-If we only want to save the shortest book:
+If we only want to view the shortest book:
```
gvwilson, line 1331: Good catch.

py-rse/bash-basics.Rmd
```diff
@@ -1324,12 +1328,17 @@ $ wc -l *.txt | sort -n | head -n 3
...
 
 ```shell
-$ wc -l *.txt | sort -n | head -n 1 > shortest.txt
+$ wc -l *.txt | sort -n | head -n 1
 ```
 
+If we were so inclined,
```
gvwilson, line 1337: I'd removed this line and start with "We could also"

py-rse/correct.Rmd
```diff
@@ -11,7 +11,30 @@ But how can we be sure?
 The short is answer is that we can't be completely certain,
 but that as in science,
 we can test our expectations against reality to help us decide if we are sure enough.
-This chapter explores ways of doing that.
+
+We'll start with our Zipf's Law project structure as continued from the previous chapter:
```
gvwilson, line 15: "as it was after" rather than "as continued from" maybe?

py-rse/package-py.Rmd
```diff
@@ -45,20 +70,11 @@ pkg_name
...
-│   └── plotcounts.py
-├── README.md
-└── setup.py
-```
+which in our case would mean changing the name of `bin/` to `zipfs_law/`.
 
-It may initially be confusing to have two folders with the same name,
+It may initially be confusing to have two folders with the same name
+(the original project directory,
+and the renamed `bin/` containing `*.py`),
```
gvwilson, line 77: removed "containing '*.py'" maybe - I don't think it adds?

py-rse/style.Rmd
```diff
@@ -11,9 +11,34 @@ This is particularly true when we look at programming style.
 Computers don't need to understand programs in order to execute them,
 but people do if they are to create, debug, and extend them.
 
-In the preceding chapters we wrote code to analyze word counts in classic novels.
+In the preceding chapters we wrote code to analyze word counts in classic novels,
```
gvwilson, line 14: no need for comma at end of line. (holy nitpicking, greg)

DamienIrving:
@k8hertweck Thanks for your work on this.

Couple of quick things:
- I've created #372 to capture your note about the fact that the Zipf's Law narrative is yet to be incorporated into the correctness and CI chapters.
- Do you think you have now resolved #344? And do we need a new issue to use `zipfs_law` instead of `zipfs-law` consistently? (i.e. I think we need to go with an underscore for the reasons outlined in the packaging chapter)

py-rse/bash-basics.Rmd
k8hertweck: TIL...this.

k8hertweck:
@DamienIrving Thanks for creating that issue! 

I'll go ahead and make updates to change the project name to underscores.

I also closed out that issue you mentioned; I'll review others and see if there is anything else reconciled by this PR.

I'll finish up mods from Greg's comments this afternoon.

py-rse/correct.Rmd
k8hertweck: Agreed, and I also checked the other chapters to ensure the language is consistent.

py-rse/package-py.Rmd
k8hertweck: yes!

py-rse/style.Rmd
k8hertweck: the more nitpicking now, the less later...

k8hertweck:
I polished off a few formatting things (in addition to the suggested changes).

PR #369, opened 2020-04-27
Clarify the constant of proportionality, closes #367

### LINKED ISSUE #367: Check intercepts in Zipf's Law plotting code
I'm worried the intercept this implies:
```
yvals = max_rank * (xvals**(-beta + 1))
```
isn't quite right. Opening this issue for me to check, but not hold up the merging of #349

cwickham:
Turns out the code was fine...just some re-wording to clarify the constant of proportionality.

### DIFF
--- a/py-rse/git-advanced.Rmd
+++ b/py-rse/git-advanced.Rmd
@@ -229,14 +229,15 @@ linear relationship.  To be specific, if Zipf's Law holds,
 thinking about rank as explained by frequency, we have \[
 r = cf^{\frac{-1}{\alpha}}
 \]
+where \(c\) is a constant of proportionality.
 The linear relationship between the log word frequency and 
 log rank is then \[
-\log(r) = C - \frac{1}{\alpha}\log(f)
+\log(r) = \log(c) - \frac{1}{\alpha}\log(f)
 \]
 This suggests that on our log-log plot 
 the points should fall on a straight line 
 with a slope of \(- \tfrac{1}{\alpha}\) 
-and intercept \(C\) (we'll see later that \(C\) is completely defined once we know \(\alpha\)).
+and intercept \(\log(c)\) (we'll see later that \(c\) is completely defined).
 Our goal is to estimate the value of \(\alpha\).
 
 #### Estimating alpha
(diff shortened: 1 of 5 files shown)


py-rse/git-advanced.Rmd
```diff
@@ -245,13 +246,16 @@ In order to determine the best method for estimating \(\alpha\) and fitting a li
...
 viewed as a function of the parameters in the statistical model that we assume generated it. 
 In maximum likelihood estimation, we estimate the parameters in the model 
 by choosing them to maximize this likelihood, although
 computationally, it is often easier to minimize the negative log likelihood function.
 @Moreno2016 define the likelihood using a parameter, \(\beta\), 
-that we will see below is directly related to the \(\alpha\) parameter in our definition of Zipf's Law.
+which is directly related to the \(\alpha\) parameter in our definition of Zipf's Law,
+through \(\alpha = \tfrac{1}{\beta-1}\).
+Under their model, the value of $c$ is the total number of unique words, or 
+equivalently the largest value of the rank.
```
DamienIrving, line 258: For this hypothetical word count distribution,
```
the: 5 times (rank 1)
at: 4 times (rank 2)
hello: 1 time (rank 3)
goodbye: 1 time (rank 3)
```
the largest value of the rank is 3 but the total number of unique words is 4. This isn't a big issue for large texts (the difference between the max rank and number of unique words will be relatively small) but to be correct should be use the number of unique words and not the maximum rank in our code?

py-rse/git-advanced.Rmd
```diff
@@ -308,8 +312,8 @@ def plot_fit(curve_xmin, curve_xmax, max_rank, beta, ax):
     ax.loglog(xvals, yvals, color='grey')
 ```
 
-where the maximum word frequency rank corresponds to intercept
-in our linear form, and \(-\beta + 1\) the slope.
+where the maximum word frequency rank corresponds to \(c\), 
+and \(-\beta + 1\) the exponent in the power law.
```
DamienIrving, line 316: Is this where you were suggesting in the previous PR that we should use -1 / alpha instead of -beta + 1?

py-rse/git-advanced.Rmd
cwickham: Depends on your definition of rank.  In R, your definition is the min rank:
``` r
rank(dplyr::desc(c(5, 4, 1, 1)), ties.method = "min")
#> [1] 1 2 3 3
```

We should be using the max rank:
```r
rank(dplyr::desc(c(5, 4, 1, 1)), ties.method = "max")
#> [1] 1 2 4 4
```
With max rank, the maximum max rank is always the same as number of values to rank.

I'm not sure what is the default in python, but we need to check.

py-rse/git-advanced.Rmd
cwickham: Yeah, that's what I was thinking.  My python is bad enough that I wasn't that confident I'd make all the right changes though...

py-rse/git-advanced.Rmd
DamienIrving: By default pandas was using the average rank, so I've changed it to max and updated the various scripts and chapters accordingly

py-rse/git-advanced.Rmd
DamienIrving: Cool - I've gone through and switched everything to alpha.

Issue #372, opened 2020-05-11
Integrate the Zipf's Law narrative into correctness and CI chapters


DamienIrving:
One section of the book where the Zipf's Law narrative is yet to be included is the chapters about testing. [narrative.md](https://github.com/merely-useful/merely-useful.github.io/blob/book/py-rse/narrative.md) suggests the following updates: 

**Chapter 14: Correctness**
This chapter introduces testing and needs to be updated to write tests for the Zipf's Law code. For example, `countwords.py` could be tested using an input text file where the count for each word is already known, and `plotcounts.py` could be given an input csv file where the alpha value in our line fitting algorithm is exactly 1 (i.e. perfect conformity to Zipf's Law).

**Chapter 15: Continuous integration**
This chapter needs to be updated so the tests developed in the previous chapter are implemented using Travis CI.

DamienIrving:
@cwickham We talk about the alpha parameter in the advanced git chapter as follows:
> "The most frequent word will occur approximately \(2^\alpha\) times as often as the second most frequent word, \(3^\alpha\) times as often as the third most frequent word, and so on."

Based on that understanding of alpha, my test for our calculation of alpha (which I'm looking to add to the correctness chapter) currently looks as follows:

```
import numpy as np
import plotcounts

def test_alpha():
    """Test the calculation of the alpha parameter."""
       
    word_counts = np.array([60, 30, 20, 15, 12, 10])
    alpha = plotcounts.get_power_law_params(word_counts)
    assert alpha == 1.0
```

The problem is that alpha = 3.03 in this example. I'm confused 😕

cwickham:
@DamienIrving, the problem is that the values you generated are consistent with the empirical law:
> "The most frequent word will occur approximately (2^\alpha) times as often as the second most frequent word, (3^\alpha) times as often as the third most frequent word, and so on."

but not statistical model that the maximum likelihood estimates are based on.  So, for instance, if alpha = 1, and the maximum word frequency is 60, we would expect to see a number of words with frequencies below 10.  How many exactly?  Well, that's complicated because the statistical model we are using (after Moreno et al), only has the behavior of the empirical law "in the tails" - Moreno's terminology, which means for large word frequencies.  

One way to get data consistent with the statistical model is to use the model to generate a set of word frequencies that are equally likely.  By adding the constraint that the maximum frequency is 60, I get (sorry about the R code):
```r
[...]

PR #379, opened 2020-05-20
Adding Zipf's narrative to correctness and CI chapters

DamienIrving:
See #372 for background.


DamienIrving:
The updated testing and continuous integration content (now merged into one chapter) with Zipf's Law narrative included is now ready for review. :smile:

I think the merged chapter should be called "testing" as opposed to "correctness", but I'll leave that proposed name change for another time.

PR #383, opened 2020-05-25
Edits to Bash basics chapter

gvwilson:
Mostly wording, but makes the pipeline command examples consistent.

Please fill in the FIXME markers as part of the review.

Requires #382.


py-rse/bash-basics.Rmd
```diff
@@ -252,10 +262,10 @@ $ ls -F /Users
 Amira/  Jun/    Sami/
```
k8hertweck, line 262: Can we make these lowercase throughout the book (to match what's written in the narrative up above)? Not sure how other chapters list them

py-rse/bash-basics.Rmd
```diff
@@ -936,9 +950,9 @@ wc: data: read: Is a directory
 ```
 
 Instead,
-we can use [wildcards][wildcard] to select a set of filenames at once.
-The most commonly-used wildcard is `*`;
-it matches zero or more characters,
+we can use [wildcards][wildcard] to specify a set of files at once.
+The most commonly-used wildcard is `*` (a single asterisk).
+It matches zero or more characters,
 so `data/*.csv` matches all of the files in the `data` directory:
```
cwickham, line 956:
```suggestion
so `data/*.txt` matches all of the text files in the `data` directory:
```

py-rse/bash-basics.Rmd
```diff
@@ -1315,13 +1330,16 @@ The vertical bar `|` between the `wc` and `sort` commands
...
+FIXME: diagram of pipes
+
 We can use `|` to build pipes of any length.
 For example,
-we can use the command `head` to get just the first three lines of sorted data, the three shortest books:
+we can use the command `head` to get just the first three lines of sorted data,
+which shows us the three shortest books:
 
 ```shell
 $ wc -l *.txt | sort -n | head -n 3
```
cwickham, line 1345: This is the first time our learners see an option that also requires an argument.  I.e. given the earlier discussion about options and arguments, is it worth pointing out that the `3` is actually being given to the `-n` option, not as input to `head`.

py-rse/bash-basics.Rmd
```diff
@@ -1798,61 +1794,68 @@ wc -l *.txt | sort -n | head -n 3
   13028 sense_and_sensibility.txt
 ```
 
-The shell prints the command it is going to run to standard error rather than to standard output,
-so that (for example) `!572 > results.txt` puts the command's output into a file
-*without* also writing the command.
+The shell prints the command it is going to re-run to standard error
+before executing it.
+This means that (for example) `!572 > results.txt`
+will the command's output into a file
```
cwickham, line 1800:
```suggestion
will direct the command's output into a file
```

py-rse/bash-basics.Rmd
```diff
@@ -2465,6 +2473,19 @@ for books in *.txt
 > done
```
cwickham, line 2473: Previous line, 2472, `$datafile` should be `$books`

k8hertweck:
The points @cwickham makes about language and motivation are really important! Additionally, I've spent multiple days this week struggling with upgrading MacOSX to Catalina, for which bash is NOT the default shell (it's now zsh). We can expect a sizable (and growing) portion of the audience will be in the same position, so I think we definitely need to clarify what it means to run bash, and what to do if you have % instead of $ in your command prompt.

py-rse/bash-basics.Rmd
gvwilson: good catch - fixed.

py-rse/bash-basics.Rmd
```diff
@@ -128,10 +137,11 @@ $ pwd
 Here,
 the computer's response is `/Users/amira`,
 which tells us that we are in a directory called `amira` that is contained in a top-level directory called `User`.
```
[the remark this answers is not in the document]
gvwilson: Good catch - thanks.

py-rse/bash-basics.Rmd
```diff
@@ -277,10 +287,11 @@ Amira   Jun     Sami
...
-> and try to avoid these shortcuts when writing code you intend other people to use.
+> Some people think this is convenient;
+> others (including us) believe it is confusing,
+> so it's best to avoid doing this.
 
-## How can I move around in the shell? {#py-rse-bash-basics-navigate}
+## Moving Around {#py-rse-bash-basics-navigate}
 
 Let's run `ls` again to look at what's in our current working directory,
 which is what the command shows if we don't ask it to show us anything else:
```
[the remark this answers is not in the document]
gvwilson: :+1:

py-rse/bash-basics.Rmd
```diff
@@ -451,19 +461,20 @@ but we'll see some uses for it soon.
 > This command is synonymous with the previous example.
 > While you may see commands written like this,
 > we don't recommend you use this approach in your own work.
-> This is because some commands take long-name options,
-> and it's very easy to mistake `-no` (meaning `-n -o`)
-> with `--no` (meaning "answer 'no' to all questions").
+> This is because some commands take [long options][long-option]
```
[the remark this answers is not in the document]
gvwilson: Added to `gloss.md` - thanks.

py-rse/bash-basics.Rmd
```diff
@@ -553,7 +564,7 @@ $ mkdir analysis
```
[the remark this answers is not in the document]
gvwilson: oops - fixed.

py-rse/bash-basics.Rmd
gvwilson: good catch - thanks

py-rse/bash-basics.Rmd
gvwilson: Good catch - paragraph added.

py-rse/bash-basics.Rmd
gvwilson: :+1:

gvwilson:
Fixes made - thanks very much for the reviews.

py-rse/bash-basics.Rmd
```diff
@@ -147,60 +157,60 @@ and `Users` (where users' personal directories are located).
...
 
 Underneath `/Users`,
 we find one directory for each user with an account on this machine.
 Jun's files are stored in `/Users/jun`,
 Sami's in `/Users/sami`,
 and Amira's in `/Users/amira`.
 This is where the name "home directory" comes from:
-when we log in as a particular user,
-the shell starts us off in the directory that holds that user's files.
+when we first log,
```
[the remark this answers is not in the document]
gvwilson: fixed

PR #398, opened 2020-06-20
Revisions to configuration chapter

gvwilson:
Ready for review.


py-rse/configuration.Rmd
```diff
@@ -128,26 +134,30 @@ mpl.matplotlib_fname()
...
 axes.labelsize       : x-large  ## fontsize of the x any y labels
 xtick.labelsize      : large    ## fontsize of the tick labels
 ytick.labelsize      : large    ## fontsize of the tick labels
 ```
-and then re-generate the Jane Eyre plot with bigger labels:
+and then re-generate the Jane Eyre plot with bigger labels
+(Figure \@ref(fig:configuration-jane-eyre-big-labels)):
 
 ```python
 $ python plotcounts.py jane_eyre.csv --outfile jane_eyre.png
```
DamienIrving, line 163:
```suggestion
$ python bin/plotcounts.py data/jane_eyre.csv --outfile results/jane_eyre.png
```

py-rse/configuration.Rmd
```diff
@@ -128,26 +134,30 @@ mpl.matplotlib_fname()
...
+as is the size of the tick labels:
 
 ```yaml
 #axes.labelsize       : medium  ## fontsize of the x any y labels
 #xtick.labelsize      : medium  ## fontsize of the tick labels
 #ytick.labelsize      : medium  ## fontsize of the tick labels
 ```
 
-We could uncomment those lines, change the sizes,
+We can uncomment those lines change the sizes to "large" and "extra large":
```
DamienIrving, line 152:
```suggestion
We can uncomment those lines and change the sizes to "large" and "extra large":
```

zipf/bin/plotcounts.py
```diff
@@ -76,15 +76,18 @@ def main(args):
     ax = df.plot.scatter(x='word_frequency', y='rank', loglog=True,
                          figsize=[12, 6], grid=True, xlim=args.xlim)
 
-    alpha = get_power_law_params(df['word_frequency'].to_numpy())
+    alpha, beta = get_power_law_params(df['word_frequency'].to_numpy())
```
DamienIrving, line 79:
```suggestion
    alpha = get_power_law_params(df['word_frequency'].to_numpy())
```
In older versions of `plotcounts.py`, `get_power_law_params` returned both `alpha` and `beta`, but now it only returns `alpha`.

zipf/bin/plotcounts.py
```diff
@@ -76,15 +76,18 @@ def main(args):
...
+
     # Since the ranks are already sorted, we can take the last one instead of
     # computing which row has the highest rank
     max_rank = df['rank'].to_numpy()[-1]
+
     # Use the range of the data as the boundaries when drawing the power law curve
     curve_xmin = df['word_frequency'].min()
     curve_xmax = df['word_frequency'].max()
 
+    # Plot the result
```
DamienIrving, line 90: Should we be encouraging comments in Python scripts? It's better for the script to be readable without comments, right?

py-rse/configuration.Rmd
```diff
@@ -425,26 +397,56 @@ if __name__ == '__main__':
...
-to multiple users and/or a large number of configuration options. 
+Programs are only useful if they can be controlled,
+and work is only reproducible if those controls are explicit and shareable.
+If the number of controls needed is small,
+adding command-line options to programs and setting those options in Makefiles
+is usually the best solution.
+As the number of options grows,
+so too does the value of putting options in files of their own.
+And if we need to manage differences between different installations of the software,
+user-specific and system-wide configuration files are probably the best choice.
```
DamienIrving, line 416: I'm a little unclear on what this means? (i.e. I read this sentence and I'm still unclear about what is a good use case for user-specific and system-wide configurations)

py-rse/configuration.Rmd
```diff
@@ -157,21 +167,25 @@ $ python plotcounts.py jane_eyre.csv --outfile jane_eyre.png
...
-### Create a matplotlib style sheet
+A global options file *is* useful, though.
+If we are using Matplotlib with [LaTeX][latex] to generate reports
+and the latter is installed in an unusual place on our computing cluster,
+a one-line change in `matplotlibrc` can prevent a lot of failed jobs.
 
-Instead of directly editing the `matplotlibrc` file to change
-the default plot characteristics,
-there are a number of pre-defined styles provided by Matplotlib:
+## Creating a Style Sheet {#configuration-stylesheet}
```
DamienIrving, line 186: Would this heading be better phrased as something like "The User Configuration File" (i.e. the general concept we are teaching) rather than "Creating a Style Sheet" (i.e. something specific to matplotlib)?

My thinking is that in this chapter we basically cover four program configuration options:
1. system-wide/global configuration files
2. user-specific configuration files
3. job-specific configuration files
4. command line options

The other headings (The Global Configuration File, Adding Command-Line Options and
A Job Control File) refer to one of those general options/concepts rather than the specific feature of matplotlib that we're using.

py-rse/configuration.Rmd
```diff
@@ -86,38 +85,45 @@ fonts:
...
+to change the size of the labels:
 
--  Edit the system-wide matplotlib configuration YAML file known as the `matplotlibrc` file
--  Create a user-specific matplotlib style sheet (which also uses the YAML format) 
--  Create a job-specific YAML file that the `plotcounts.py` script can parse
--  Add some new command-line options to `plotcounts.py` 
+-  Edit the system-wide Matplotlib configuration file
+   (which would affect everyone using this computer).
+-  Create a user-specific Matplotlib style sheet.
+-  Create a configuration file to set plotting options in `plotcounts.py`.
```
DamienIrving, line 115:
```suggestion
-  Create a job-specific configuration file to set plotting options in `plotcounts.py`.
```

DamienIrving:
One more thought on a configuration topic that we might want to consider... at the moment we `import utilities` (i.e. the module we wrote) in `countwords.py`. That works fine if we run `countwords.py` from the `zipf/bin` directory (or a directory higher up the tree) but won't work if we execute from `zipf/results` or some other random directory on our system. Do we need to talk about `sys.path` and how we tell Python where to look for our modules?

zipf/bin/plotcounts.py
gvwilson: I've never bought into that 100%, and I think it helps orient readers in this case.

py-rse/configuration.Rmd
gvwilson: Changed the wording to emphasize that this is for sys admins to hide differences on multi-user setups (e.g. research clusters).

py-rse/configuration.Rmd
gvwilson: Good catch - title changed.

gvwilson:
The import works on my machine: in general, `import x` will always find an `x.py` that's in the same directory as the file doing the import, even if that directory is not on `PYTHON_PATH`.

PR #400, opened 2020-06-20
Edits to error-handling chapter

gvwilson:
Depends on #398.

PR #401, opened 2020-06-22
Teams chapter

gvwilson:
Depends on #400 - please only review changes in `py-rse/teams.Rmd`.


py-rse/teams.Rmd
```diff
@@ -318,91 +366,70 @@ Every issue can hold:
...
 6.  Describe each problem separately so that each one can be tackled on its own.
     This parallels the rule about creating a branch in version control for each bug fix or feature
     discussed in Section \@ref(git-advanced).
 
-Here is an example of a well-written bug report with all of the components mentioned above,
-as well as the common features of all issues:
+Here is an example of a well-written bug report with all of the components mentioned above:
 
 ```text
 ID: #25
```
DamienIrving, line 435: The creator and owner in his example bug report could be Sami and Amira to fit with our narrative?

Also, the current example bug report uses a hypothetical script `wordbase.py`, which is somewhat confusing because it sounds like one of the Zipf's scripts (`countwords.py`) but isn't. It might be better to submit a genuine bug report? For instance, `countwords.py` doesn't handle em-dashes properly. Instead of splitting on an em-dash, it counts the whole thing as a word. e.g.
```
$ python bin/countwords.py data/dracula.txt
...
time--which,1
sun--now,1
tops--and,1
off--in,1
before--came,1
... 
```
(That's just a few of the many em-dash outputs when you run `countwords.py` on Dracula.)

DamienIrving: Once we've created a `zipf` directory in Amira's repo we could also put the text from this bug report into an actual issue at that repo and screenshot it?
https://github.com/amirakhan

py-rse/teams.Rmd
```diff
@@ -445,17 +468,22 @@ You can't force new issues to be as complete as you might like,
...
+to make things easier to search and organize.
 Labels are also often called [tags][tag];
 whatever term is used,
-they are just a word or two to help users sort and search among tasks.
+each one is just a descriptive word or two.
 
-GitHub allows project owners to create any labels they want
-that users can then add to their issues.
-A small project should always use some version of these three labels:
+GitHub allows project owners to define any labels they want.
```
DamienIrving, line 485: Once we've created a bug report/issue in Amira's repo (see previous comment) we could screenshot the drop down labels menu to illustrate what we're talking about in this labels section? e.g. 
Screen Shot 2020-06-23 at 1 41 02 pm

py-rse/teams.Rmd
gvwilson: 1. Changed bug report - good catch.
2. Filed issue #405 about screenshot.

py-rse/teams.Rmd
gvwilson: Added to issue #405.

py-rse/teams.Rmd
```diff
@@ -445,17 +465,22 @@ You can't force new issues to be as complete as you might like,
 but you can use an issue template to make it easier for contributors to
 remember and complete documentation about bug reports.
 
-## How can we use labels to organize work? {#teams-labels}
+FIXME: how to respond to a bug report.
```
cwickham, line 468: This seems well addressed by the following paragraph. Can the FIXME be removed? Or did you have something else in mind?

py-rse/teams.Rmd
```diff
@@ -33,131 +39,158 @@ The project directory should now include:
...
+More importantly,
+it is the right thing to do.
+
+> **Terminology**
+>
+> [Privilege][privilege] is an unearned advantage given to some people but not all,
+> while [oppression][oppression] is systemic inequality that benefits the privileged
+> and harms those without privilege @Auro2019.
+> In Europe, Australia, and the Americas,
+> a straight, white, affluent, physically able male
```
cwickham, line 73: I was going to suggest "Australasia", but [then read](https://en.wikipedia.org/wiki/Australasia) that that means different things to Australians and New Zealanders.  I think we could add NZ to the list...

py-rse/teams.Rmd
gvwilson: If you're happy, I'm happy :-)

py-rse/teams.Rmd
gvwilson: amended

PR #419, opened 2020-07-10
Minor fixes

DamienIrving:
The editing pass introduced a few inconsistencies that this PR fixes:
- Updates to the word count function meant that the word count distribution for Dracula (which features in many places, including in the testing chapter) had changed
- The example assertion that was added to `get_power_law_params` evaluated true no matter what (`SyntaxWarning: assertion is always true, perhaps remove parentheses?`) and the solution was a little too complex so I went for a simpler assertion
- The data for testing needs to be kept separate from the actual data or else it gets included when you run `make all` (from the automation chapter)


gvwilson:
Good catch - thank you.

PR #425, opened 2020-07-22
Review of reordering 

cwickham:
Using this PR for any changes I can make directly, otherwise I'm opening issues.

### DIFF
--- a/py-rse-index.Rmd
+++ b/py-rse-index.Rmd
@@ -158,7 +158,8 @@ that are freely available from [Project Gutenberg][project-gutenberg].
 Appendix \@ref(install) explains
 how to download the novels and other files you will need.
 When you are done,
-you should have a directory (also called a [folder][folder])
+you should have a directory (also called a [folder][folder]) 
+called `zipf`,
 containing a single sub-directory called `data`
 with the following contents:
 
--- a/py-rse-index.Rmd
+++ b/py-rse-index.Rmd
@@ -179,16 +179,16 @@ we will show you how to:
 
 -   Use the Unix shell to efficiently manage your data and code.
 -   Write Python programs that can be run at the command line.
+-   Write and review code to make it readable as well as correct.
 -   Use Git and GitHub to track and share your work.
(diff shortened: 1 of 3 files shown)


cwickham:
Overall, the reorganization has been really successful!  There are only two chapters I thought had some issues:

* `style.Rmd` I think this is in the right location, right after scripting, but it still feels a bit out of place.  I think mostly because it doesn't advance the narrative. We could just be up front about this in the introduction to the chapter e.g. "We wrote the scripts so far to follow good style, so the examples used in this chapter aren't related to the zipf project".  

* `project.Rmd` needs splitting/merging (as also pointed out in a few comments and FIXMEs). I'll open a separate issue with some suggestions.

DamienIrving:
I guess there's minor edits we could make to `style.Rmd` to work in the narrative a bit (e.g. we could run `pycodestyle` on our scripts, but the output is pretty boring - see below), but in general it's probably not a good idea to write sub-optimal code in the scripting chapter only to change it all in `style.Rmd`. I agree that `style.Rmd` is in the right place if we want to include it as a regular chapter, but I'd also be open to the idea of it being some kind of appendix that essentially explains the coding conventions we followed throughout the book.
```
$ pycodestyle collate.py 
collate.py:18:80: E501 line too long (83 > 79 characters)
collate.py:26:80: E501 line too long (80 > 79 characters)

$ pycodestyle countwords.py 
countwords.py:1:80: E501 line too long (81 > 79 characters)

$ pycodestyle plotcounts.py 
plotcounts.py:12:80: E501 line too long (88 > 79 characters)
plotcounts.py:76:80: E501 line too long (80 > 79 characters)
plotcounts.py:88:80: E501 line too long (82 > 79 characters)
plotcounts.py:103:80: E501 line too long (80 > 79 characters)
```

gvwilson:
If `style.Rmd` doesn't fit the narrative, let's move it to an appendix - we can tell people up front that we're following good style from the start, but since a lot of people don't, we've described them with examples.

Issue #432, opened 2020-07-24
Proposal: Add a completed project listing as an Appendix


cwickham:
## Proposal

Include the complete tree for `zipf` as an Appendix, with each file linked to the chapter/section it is introduced.  

## Background

@DamienIrving provided a great complete tree for `zipf` in #430.  The complete tree is pretty complicated, so I didn't include the full thing in the re-working of project structure #431.  However, it would be nice to be able to link to the complete tree, and it also provides a useful way for a reader to navigate the book.

## Pros and Cons

**Pro:**
-   Adds an additional way to navigate text
-   Not too much work to implement beyond what Damien has already provided

**Con:**
-   One more thing that needs to stay updated if the content of the project changes

## Alternatives

Do nothing in the book, and use the GitHub account for our characters to perform this service.

k8hertweck:
I really like this idea; the truncated `tree` output at the beginning of each chapter provides minimum info, but a specific place to look for the complete project would be great (and something I wish we'd had all along!)

DamienIrving:
I like this idea too. For now, #434 contains a new file `py-rse-final-tree.md` which shows the whole tree (I missed a few files in https://github.com/merely-useful/merely-useful.github.io/issues/430#issuecomment-663283957). We could convert that file to an appendix before or after that PR is merged.

Issue #412, opened 2020-07-01
Add to error handling chapter


gvwilson:
From @DamienIrving's comments on #400 - this needs to be done and rippled through subsequent chapters, but @gvwilson doesn't want to hold up merges.

We might want to fill out `py-rse/solutions.Rmd` (and update `collate.py` accordingly) for this chapter before merging. A bunch of new concepts about catching errors and logging are introduced during the chapter and we talk a little about how they might be introduced into `collate.py`, but we don't actually go ahead and put those changes into `collate.py` and run it. That's left up to the exercises. As such, it's not clear what the final version of `collate.py` (i.e. the one that is later packaged up so that it can be pip installed) actually looks like.

I'm thinking it needs to look like this:

```
"""Combine multiple word count CSV-files into a single cumulative count."""
import csv
import argparse
from collections import Counter
import logging
import utilities

def update_counts(reader, word_counts):
    """Update word counts with data from another reader/file."""
    for word, count in csv.reader(reader):
        word_counts[word] += int(count)

def main(args):
    """Run the command line program."""
    log_level = logging.DEBUG if args.verbose else logging.WARNING
    logging.basicConfig(level=log_level, filename=args.logfile)
    word_counts = Counter()
    logging.info('Processing files...')
    for file_name in args.infiles:
        logging.debug(f'Reading in {file_name}...')
        if file_name[-4:] != '.csv':
            raise OSError(f'{file_name} must end in `.csv`.')
        with open(file_name, 'r') as reader:
            logging.debug('Computing word counts...')
            update_counts(reader, word_counts)
    utilities.collection_to_csv(word_counts, num=args.num)

if __name__ == '__main__':
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument('infiles', type=str, nargs='*', help='Input file names')
    parser.add_argument('-n', '--num', type=int, default=None,
                        help='Limit output to N most frequent words')
    parser.add_argument('-v', '--verbose', action="store_true", default=False,
                        help="Change logging threshold from WARNING to DEBUG")
    parser.add_argument('-l', '--logfile', type=str, default='collate.log',
                        help='Name of the log file')
    args = parser.parse_args()
    main(args)
```
We can then run it as follows:
```
$ python bin/collate.py results/dracula.csv results/moby_dick.csv -n 10 -v 
the,22559
and,12306
of,10446
to,9192
a,7629
in,6745
i,6557
that,5373
it,4464
he,4260

[...]

PR #442, opened 2020-08-18
Adding an appendix for the final directory tree

DamienIrving:
This PR primarily addresses #432 by creating a new appendix that shows the final directory tree. It also lists the Chapter/Section/Exercise that each file was created and any subsequent Chapter/Section/Exercise that it was modified.

It also addresses #412 (relating to the final version of `collate.py`) because I worked on some exercises/solutions while figuring out which exercises modify which files in the tree.

There are also some minor changes to other files wherever I noticed inconsistencies between those files and the final directory tree.


py-rse/solutions.Rmd
```diff
@@ -693,6 +704,26 @@ FIXME
...
+Thank you for your interest in contributing to the Zipf's Law package!
+
+If you are new to the package and/or collaborative code development on GitHub,
+feel free to discuss any suggested changes via issue or email.
+We can then walk you through the pull request process if need be.
+As the project grows,
+we intend to develop more detailed guidelines for submitting
+bug reports and feature requests.
+
+We also have a code of conduct (see `CONDUCT.md`).
```
lwjohnst86, line 723:
```suggestion
We also have a code of conduct (see [`CONDUCT.md`](CONDUCT.md)).
```

Just so it links directly to the file, rather than having to find it.

py-rse/tree.Rmd
```diff
@@ -0,0 +1,144 @@
...
+    ├── plotcounts.py
+    ├── plotparams.yml
+    ├── test_zipfs.py
+    └── utilities.py
+```
+
+Each file was introduced and subsequently modified
+in the following chapters, sections and exercises:
+
+`.gitignore`: Introduced in Section \@ref(git-cmdline-ignore).
```
lwjohnst86, line 67: Would this be better as a list? Or as a table?

lwjohnst86 (approved):
Looks good! I can't comment on the python code, but the exercises seem good. The other content re the tree looks great! A couple minor comments.

py-rse/tree.Rmd
DamienIrving: It would possibly be better as a table, but I figured I'd leave it like this at the moment and the book editors can indicate how best to present the information.

PR #449, opened 2020-08-26
Small fixes to ensure narrative/tree consistency

DamienIrving:
I found more minor inconsistencies in the handling of the file tree through the book.

### DIFF
--- a/py-rse/automate.Rmd
+++ b/py-rse/automate.Rmd
@@ -30,6 +30,7 @@ zipf/
 │   ├── collate.py
 │   ├── countwords.py
 │   ├── plotcounts.py
+│   ├── script_template.py
 │   └── utilities.py
 ├── data
 │   ├── README.md
--- a/py-rse/bash-advanced.Rmd
+++ b/py-rse/bash-advanced.Rmd
@@ -41,10 +41,13 @@ For historical reasons
 a file full of shell commands is usually called a [shell script][shell_script],
 but it is really just another kind of program.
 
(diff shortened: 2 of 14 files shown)


k8hertweck:
I'm to blame for the python template script, shell script, and summary outputs not being included in the tree output originally; I was being a bit lazy and hedging my bets that they weren't mission critical and we didn't need to account for them as thoroughly. Thank you for clarifying it, and also checking for consistency in paths for running the scripts!

I was just working on the images for the git chapters, by recreating the project in [Amira's GitHub](https://github.com/amira-khan/zipf) as a part of #440 . The current project doesn't have `bin/script_template.py`, which doesn't affect the figure itself, but does alter the project (if anyone tries to reference it online). I'll check back tomorrow to confirm someone else (less tired than me) has reviewed and merged this PR, and I'll finish the figures accordingly.

DamienIrving:
@k8hertweck No problem at all. I'm a bit on the fence about whether we should include `script_template.py` in the repo. We remove the summary outputs at the end of the advanced command line chapter because we aren't going to be using them anymore, so we could do a similar thing with `script_template.py`? We don't write any new scripts for the rest of the book, so we don't need it going forward...

lwjohnst86 (approved):
Looks great! No comments

k8hertweck:
Since this has been merged already, I'll go ahead and redo Amira's repo (and hope we don't find anything else that needs to change!)

PR #452, opened 2020-08-26
Solutions for Git Cmdline exercises

lwjohnst86:
Plus some clarifications.

The last exercise, item 2, I don't see what the `^` is so didn't answer.


chapters/solutions.Rmd
```diff
@@ -514,11 +514,25 @@ or both.
...
+    nothing added to commit but untracked files present (use "git add" to track)
+    ```
+
+- Nothing has happened to the file, it still exists but Git no longer has it in
+the staging area.
 
 ### Exercise \@ref(git-cmdline-ex-word-diff) {-}
 
-FIXME
+Using `git diff --word-diff` might be easier since it shows exact what has been
```
DamienIrving:
```suggestion
Using `git diff --word-diff` might be easier since it shows exactly what has been
```

chapters/solutions.Rmd
```diff
@@ -677,15 +691,25 @@ ENIAC was the world's first general-purpose electronic computer.
 
 ### Exercise \@ref(git-cmdline-ex-diff) {-}
 
-FIXME: solution for exercise on git diff
+1. `git diff HEAD~9 bin/plotcounts.py` compares what has changed between the
```
DamienIrving, line 694: I don't really understand the exercise associated with this solution, which currently reads as follows:
1.   What will the command `git diff HEAD~9 bin/plotcounts.py` do if we run it?
2.  What does it actually do?

What is the difference between "what the command does if we run it" and "what it actually does"? It's basically asking the same thing twice.

Also, we might want to use `HEAD~2` instead of `HEAD~9`. At this point of the book there have been only four commits to the repo, and only two of them (the initial commit and then the second to last commit) involve `plotcounts.py`:

1.  `git commit -m "Add scripts, novels, word counts, and word rank plot"`
2.  `git commit -m "Add plot of word counts for 'Dracula'"`
3. `git commit -m "Edit to plot frequency against rank on log-log axes"`
4. `git commit -a -m "Update dracula plot"`

chapters/solutions.Rmd
```diff
@@ -677,15 +691,25 @@ ENIAC was the world's first general-purpose electronic computer.
...
+because the changes are in the staged area and checkout would affect
+the working directory.
 
 ### Exercise \@ref(git-cmdline-ex-blame) {-}
 
-FIXME: solution for exercise on git blame
+1. Each line of output corresponds to a line in the file, who was last
+to modify the line, when that change was made, and what the file name was
+or is called.
+2. FIXME: I don't see any ^ when I use it on a random file in the book repo.
```
DamienIrving, line 712: The ^ designates the boundary commit. (In other words, the lines of the file that have a ^ haven't changed since the initial commit.)
https://stackoverflow.com/questions/13105858/why-would-a-lines-sha-in-a-git-blame-view-have-a-leading-caret

If you run `git blame bin/plotcounts.py` at this point of the book (when there's only been one committed change to `plotcounts.py` since the initial commit) almost every line has a ^

DamienIrving (changes requested):
Thanks for doing these solutions, @lwjohnst86. I've made a few suggested changes.

chapters/solutions.Rmd
gvwilson: :+1: to your suggestions - the question was originally "What do you think it will do?" and "What does it actually do?" The trick is that there isn't enough history for `HEAD~9` to be meaningful because (as you point out) there have only been four commits.

chapters/solutions.Rmd
```diff
@@ -677,15 +691,25 @@ ENIAC was the world's first general-purpose electronic computer.
...
+2.  The previous git command takes the state of the file at each point and then
+compares them.
+3.  `git diff HEAD bin/plotcounts.py` compares what has been changed in
+`bin/plotcounts.py` with the previous commit.
 
 ### Exercise \@ref(git-cmdline-ex-unstage) {-}
 
-FIXME: solution for exercise on git unstage
+Using `git checkout` on a staged file does not unstage it. That's 
+because the changes are in the staged area and checkout would affect
```
gvwilson: "staged area" or "staging area"?

chapters/solutions.Rmd
gvwilson: good point - I don't know how to amend the exercise given the state of the repo at this point in the book, but I think `git blame` is important. Maybe move this exercise to `git-advanced` when there will be more history to explore?

lwjohnst86: I made all the other changes but I'll leave this for now. Should we move it? Let me know and I will.

PR #454, opened 2020-08-27
Figures and corrections for git-cmdline.Rmd and git-advanced.Rmd

k8hertweck:
To replace #440, since I had a pretty serious rebase FAIL. 

Still in draft, but the good news is that I also caught a few errors in the lesson text that rendered `plotcounts.py` inoperable. It's now consistent (as far as I can tell) with the copy in the `zipf` directory.

143618a includes a note that discusses git versions, since the newest versions actually suggest different commands during the general git workflow. I'm tempted to include additional pop-outs throughout these two chapters to specifically address these new commands; while the old ones seem necessary for backwards compatibility (I'm thinking of clusters folks are running with older versions), it seems disingenuous to ignore that many readers will use the materials with a fresh install, and thus see something quite different from what we're teaching.

Amira repo: https://github.com/amira-khan/zipf

Figures involved:
- [x] repo-link
- [x] repo-history
- [ ] 5.7 rework with only commands used in this book
- [ ] ~~6.1 repository state???~~
- [x] 6.3 forking
- [x] 6.4 after forking
- [x] 6.5 after sami pushes
- [x] 6.6 starting pull request
- [x] 6.7 summary of pull request
- [x] 6.8 filling in pull request
- [x] 6.9 new pull request
- [x] 6.10 viewing pull request
- [x] 6.11 listing pull requests
- [x] 6.12 pull request details
- [x] 6.13 files changed
- [x] 6.14 comment marker
- [x] 6.15 writing comment
- [x] 6.16 pull request with comment
- [x] 6.17 pull request with fix
- [x] 6.18 successful merge
- [x] 6.19 conflict in a pull request


cwickham:
Figures so far look great! My thoughts on the three remaining ones:
* **5.7** It seems like the big adds in this figure are `git push` and `git pull`. Perhaps this should just be Figure 5.3, with the addition of a Remote repo (and label for Local) and the `push` and `pull` arrows.
* **6.1** It's not clear to me what this figure will look like?  I can imagine something in the spirit of a flow diagram (like those at https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow),  but since we haven't shown anything like that, it may just be confusing.  Does this point really need a figure?
* **6.19** Guessing this just needs some time to get done...

Re: Git versions...I think this highlights that I need to update Git (or my OS 🤮 ), but I agree we should probably acknowledge the differences learners may see.

k8hertweck:
@cwickham Thanks for looking over! I finished 6.19, and also added a note about git versions (yes, it's kind of painful, but I'm rationalizing that some folks will be working on clusters that don't update git frequently, which validates why we're referencing the older version's commands?).

I feel the same way about 6.1; I didn't create one because I was having a tough time envisioning what it would look like.

I'm happy to rework 5.3 to accommodate 5.7. I figured 5.7 would have to be redone anyway since I'm not sure where it came from?

lwjohnst86:
I didn't see if this was resolved, but do we have permission to use the PhD Comics image?

k8hertweck:
> I didn't see if this was resolved, but do we have permission to use the PhD Comics image?

@gvwilson has an email from a few years back granting permission; not sure if the publisher will require something more.

PR #456, opened 2020-08-30
Exercises for configuration chapter

gvwilson:


chapters/config.Rmd
```diff
@@ -459,6 +477,31 @@ How can we make this the default for all of our plots?
     that reads settings from a file.
```
DamienIrving: I'm a little confused by "an option that writes all of its settings to a file that reads settings from a file". Would you be able to write out the solution (see comment in the solutions file below)?

chapters/solutions.Rmd
```diff
@@ -868,11 +868,77 @@ FIXME
...
+```
+
 ### Exercise \@ref(config-ex-accessible)
 
 FIXME
 
-### Exercise \@ref(config-ex-debug)
+### Exercise \@ref(config-ex-saveload)
+
+FIXME
```
DamienIrving: Would you be able to add the solution here? It needs to build on the repo version of [`plotcounts.py`](https://github.com/merely-useful/merely-useful.github.io/blob/book/zipf/bin/plotcounts.py`).

DamienIrving (changes requested):
Thanks, @gvwilson. I've added a commit to this branch with updates to the exercise and solution about updating `Makefile`, because it involved a new concept that wasn't introduced in the automation chapter (the text manipulation function `word`).

Would you be able to add solutions for the other exercises? (I added a comment about doing this for one of the exercises, but I actually think it needs to be done for all.) It's hard to review the appropriateness of a new exercise without seeing what you think the solution should be.

DamienIrving:
@gvwilson Was this PR closed by mistake? The updates I added to the exercise and solution about the `Makefile` definitely need to be merged, and if you don't want to figure out solutions for the other exercises now we could merge and come back to the solutions later?

gvwilson:
?? I thought I merged this one, including your changes ??

DamienIrving:
@gvwilson Not sure what happened. Anyway, we should fill in the solutions before merging. I'll hopefully have time to work on them this weekend.

lwjohnst86 (approved):

PR #462, opened 2020-09-03
Creating index for book

gvwilson:
1.  Add `\index{...}` to chapters.
2.  Add missing glossary entries.
3.  Add links for glossary entries - it's a real pain that we have to duplicate keys in `links.md` and `glossary.md`, but it'll get us through this book.


chapters/teams.Rmd
```diff
@@ -473,10 +477,11 @@ SOFTWARE.
 
 The MIT license, the GPL, and the Hippocratic License are intended for use with software.
 When it comes to data and reports,
-the most widely used family of licenses are those produced by [Creative Commons][creative-commons].
+the most widely used family of licenses are those produced
+by [Creative Commons][creative-commons].\index{Creative Commons}
 These have been written and checked by lawyers and are well understood by the community.
 
-The most liberal option is referred to as [CC-0][cc_license],
+The most liberal option is referred to as [CC-0][cc_license],\index{Creative Commons!CC-0 license}\index{public domain}
```
k8hertweck, line 484: My gut says to include `\index{open source license!CC-0 license}` here, and similar for CC-BY below, to make them more discoverable.

k8hertweck:
What's the scope of this PR? Only one comment about suggested tags for indexes here; do you want a more thorough review for additional tags too?

gvwilson:
Probably the most useful think would be to format it all and look through the index (~6 pages) to see if the entries themselves make sense - there will always be things in the book that should be indexed but aren't, but if any of the entries I've made are nonsensical, now's the time to fix them.

PR #465, opened 2020-09-04
Fixed broken figure links in packaging chapter

DamienIrving:
A bunch of the figures in the packaging chapter weren't displaying because the location of the html file was included as a comment in the r / knitr include graphics commands. I've removed the comments about the location of the html file and have instead re-done the screen shots so that they show the location in the address bar of the web browser.


gvwilson:
No idea why they weren't displaying for you - when I moved the back-ticked paths into comments and rebuilt they were showing up fine for me in PDF. Thanks for catching it - I'll merge this one now.

DamienIrving:
Thanks for merging, @gvwilson. The PDF build was fine - it was the website where the comments were messing things up 😄

PR #467, opened 2020-09-04
Formatting glossary entries with Lua filters to Pandoc.

gvwilson:
1.  Glossary terms in the main text were formatted as `[term][key]`, with an entries in `./links.md` of the form `[key]: glossary.html#key` so that the hyperlink would point at the glossary entries in `glossary.md`.

1.  Those links worked fine in the HTML version, but became footnotes with URLs in the PDF version (because Bookdown uses footnote links).

1.  This PR therefore replaces all Markdown-style `[term][key]` references in the main text with LaTeX-style `\gref{term}{key}` references.

1.  And removes the forwarding links from `./links.md` (which is good, since they frequently fell out of step).

1.  But Pandoc doesn't know how to handle user-defined LaTeX commands, so this PR adds a Lua filter for the HTML version and a slightly different Lua filter for the PDF version.

1.  Each Lua filter is handed a portion of the abstract syntax tree (AST) that Pandoc produces while parsing the input file and has to return a modified or replacement data structure.

1.  To find out what the AST looks like, run `pandoc -f markdown -t json -o ast.json index.Rmd`.

1.  By inspection, the filter for PDF needs to escape underscores in keys, i.e., replace "_" with "\_".

1.  By inspection, the filter for HTML needs to turn `\gref{term}{key}` into `<a href="glossary.html#key">term</a>`, and also needs to change the type of the AST element from "tex" to "html".

1.  The link-checking script `bin/links.py` was renamed to `bin/linkcheck.py` (with a corresponding change to the call in `./Makefile`) and slimmed down so that it only checks the integrity of external Markdown-style links.

1.  A new script `bin/glosscheck.py` was added (with a corresponding target in `./Makefile`) to check the integrity of glossary entries.

1.  Definitions for "ORCID" and "YAML" were added to the `glosario` glossary, `glossary-slugs.txt` was updated, and `glossary.md` was regenerated.

1.  The `\gref` command was defined in `latex/preamble.tex`.

Note: if `\gref{term}{key}` appears on a line of its own, Pandoc seems to insert a paragraph break after it, even if Markdown would format it as part of the same paragraph as the lines that follow it. No idea why, but the easy fix is to ensure that `\gref{term}{key}` is never orphaned.


lwjohnst86:
Wow, really nicely done!! I'll take a look at this tomorrow morning.

lwjohnst86 (approved):

PR #482, opened 2020-09-12
Fix solution headings so TOC looks better. 

### LINKED ISSUE #474: Chapter headings in HTML ToC for solutions are wrong
`make html` then view:

Screen Shot 2020-09-09 at 11 37 57 AM

lwjohnst86:
Resolves #474

@gvwilson is this fine? I couldn't do it with the `\@ref()` because it ends up being a Pandoc templating thing which I don't know enough about to get into. I think this actually looks better than if it was numbers :) 

screenshot

### DIFF
--- a/chapters/solutions.Rmd
+++ b/chapters/solutions.Rmd
@@ -2,7 +2,7 @@
 
 FIXME: Update the ordering of this section when we are finished rearranging.
 
-## Chapter \@ref(bash-basics)
+## Chapter: [The Basics of the Unix Shell]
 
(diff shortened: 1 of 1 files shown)


gvwilson:
I think that duplicating chapter titles is going to bite us - what about making the sections unnumbered to sidestep the problem entirely? Looking at it again, ToC entries that say "Chapter 5" aren't particularly useful or important.

lwjohnst86:
Alright, it now no longer lists them in the TOC. Only can click Solutions now.

PR #487, opened 2020-09-13
Folding in dedication

gvwilson:


includes/before_body.tex
```diff
@@ -6,7 +6,33 @@
...
+--- Luke and Joel
+
+~\\
+
+To my parents Judy and John\\
+who taught me to love books and everything I can learn from them.\\
+--- Kate\\
+
+~\\
+
```
cwickham, line 26:
```suggestion

To Joshua\\
--- Charlotte\\

~\\

```

includes/before_body.tex
```diff
@@ -6,7 +6,38 @@
 \thispagestyle{empty}
 
 \begin{center}
-FIXME
+To David Flanders\\
+who taught me so much about growing and sustaining coding communities.\\
```
DamienIrving, line 10:
```suggestion
who has taught me so much about growing and sustaining coding communities.\\
```

DamienIrving (approved):

includes/before_body.tex
gvwilson: I shortened it by one word to avoid an ugly line break (the last word was spilling over). We can put it back and worry about it in the edit?

DamienIrving: No problem - in that case you can just leave the word "has" out :smile:

PR #496, opened 2020-09-20
Solutions to configuration and packaging exercises.

gvwilson:
1.  Add solution to "save configuration" exercise.
    -   This modifies some of the `zipf` material but in a non-breaking way.
        @DamienIrving please check - do we want these changes in the main `zipf` directory
        or should they be put elsewhere?

1.  Removed the "reload configuration" part of the "save configuration" exercise:
    it turns out that `matplotlib` uses a class of its own rather than a dictionary to store parameters,
    and while it can be reconstructed, it felt like getting into the weeds.

1.  Removed the exercise on tracing where configuration parameters come from:
    it's easy to do for flat configurations but required recursion for nested parameters,
    which I think we want to steer away from.

1.  Add a new directory `data-package` with a solution to the corresponding exercise.
    This is not a full Python package, but illustrates the core idea.

### DIFF
--- a/chapters/solutions.Rmd
+++ b/chapters/solutions.Rmd
@@ -1245,6 +1245,12 @@ def main(args):
         return
     df = pd.read_csv(args.infile, header=None, names=('word', 'word_frequency'))
     # ...carry on producing plot...
+
+
+def save_configuration(filename, params):
+    """Save configuration to a file."""
+    with open(filename, 'w') as reader:
+        yaml.dump(params, reader)
 ```
 
 Finally, we add a target to `Makefile` to try out our change.
@@ -1309,6 +1315,8 @@ def set_plot_params(param_file):
 
 2.  However, Windows INI files only provide one level of sectioning,
     so complex configurations are harder to express.
+    Thus, while YAML may be a bit more difficult to get started with,
+    it will take us further.
 
 ### Exercise \@ref(config-ex-consistency) {-}
 
@@ -1710,3 +1718,41 @@ most items would be checked off of the list.
 
 Depending on how often you update your Python packages, you will have very little
 or a lot to update.
+
+### Exercise \@ref(packaging-ex-pratchett) {-}
+
+The directory tree for the `pratchett` package is:
+
+```
+pratchett
+├── pratchett
+│   └── quotes.py
+├── README.md
+└── setup.py
+```
+
+`README.md` should contain a basic description of the package and how to install/use it,
+while `setup.py` should contain:
+
+```
+from setuptools import setup
+
+
+setup(
+    name='pratchett',
+    version='0.1',
+    author='Amira Khan',
+    packages=['pratchett'],
+)
+```
+
+The following sequence of commands will create the development environment,
+activate it,
+and then install the package:
+
+```
+$ conda create -n pratchett python
+$ conda activate pratchett
+(pratchett)$ cd pratchett
+(pratchett)$ pip install -e .
+```
--- a/chapters/solutions.Rmd
+++ b/chapters/solutions.Rmd
@@ -1200,7 +1200,7 @@ parser.add_argument('--style', type=str, choices=plt.style.available,
                     default=None, help='matplotlib style')
 ```
 
-Use the style at the top of the `main function:
+Use the style at the top of the `main` function:
 ```python
 def main(args):
     """Run the command line program."""
--- a/chapters/solutions.Rmd
+++ b/chapters/solutions.Rmd
@@ -1706,56 +1706,6 @@ Depending on what was done before using the checklist, there will be either
 very little that needs to be updated or a lot. If this course was followed,
 most items would be checked off of the list.
 
-### Exercise \@ref(packaging-ex-data) {-}
-
-The central idea is that a Python file can find out what directory it was read from,
-so it can look for other files in that directory.
-This means that we don't have assume the package has been installed in a particular place
-or that some environment variable will tell us where to look for our data.
-To test this idea out:
-
-1.  Create a directory called `data-package`.
-
-1.  Put a copy of `collated.csv` in that directory.
-
-1.  Add a file called `collated.py` to that directory.
-
-    ```python
-    '''Example data package.'''
-
-    import os
-    import csv
-
-    def getData():
-        '''Load 'collated.csv' from this file's directory.'''
-        filename = os.path.join(os.path.dirname(__file__), 'collated.csv')
-        with open(filename, 'r') as reader:
-            return [row for row in csv.reader(reader)]
-    ```
-
-1.  Create another file called `test.py`:
-
-    ```python
-    import collated
-
-    data = collated.getData()
-    for row in data:
-        print(row)
-    ```
-
-1.  Run the test program.
-
-The sequence of steps is:
-
-1.  `test.py` imports `collated.py`.
-
-1.  `collated.py` uses `os.path.dirname(__file__)` to get the absolute path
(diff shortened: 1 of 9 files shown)


zipf/bin/plotcounts.py
```diff
@@ -109,5 +117,9 @@ def main(args):
                         help='YAML file containing matplotlib parameters')
     parser.add_argument('--style', type=str, nargs='*', choices=plt.style.available,
                         default=None, help='matplotlib style')
+    parser.add_argument('--saveconfig', type=str, default=None,
+                        help='Save configuration to file')
+    parser.add_argument('--loadconfig', type=str, default=None,
```
DamienIrving: We don't need the `--loadconfig` option because the existing `--plotparams` option already takes a YAML file full of matplotlib parameters.

chapters/solutions.Rmd
```diff
@@ -1165,11 +1163,76 @@ parser.add_argument('--style', type=str, nargs='*', choices=plt.style.available,
...
+                        help='Save configuration to file')
+    args = parser.parse_args()
+    main(args)
+```
+
+Next, we add three lines to `main` to act on this option *after* all of the plotting parameters have been set.
+For now we use `return` to exit from `main` as soon as the parameters have been saved;
+this lets us test our change without overwriting any of our actual plots.
+
+```python
```
DamienIrving, line 1236: In this code block where you define the updated `main` function you also need to define the new `save_configuration` function.

```
def save_configuration(filename, params):
    """Save configuration to a file."""
    with open(filename, 'w') as reader:
        yaml.dump(params, reader)
```

chapters/solutions.Rmd
```diff
@@ -1187,15 +1250,24 @@ def set_plot_params(param_file):
...
 
-TODO: Add answers to the following questions:
-
-- Which file format do you find easier to work with?
-- What other factors should influence your choice of a configuration file syntax?
+1.  Most people seem to find Windows INI files easier to write and read,
+    since it's easier to see what's a heading and what's a value.
 
+2.  However, Windows INI files only provide one level of sectioning,
+    so complex configurations are harder to express.
```
DamienIrving, line 1317: Presumably there are reasons why YAML is now more popular than INI? Maybe we should mention them in the solution, otherwise it seems a little odd that we didn't use INI.

chapters/solutions.Rmd
```diff
@@ -1582,7 +1654,53 @@ most items would be checked off of the list.
 
 ### Exercise \@ref(packaging-ex-data) {-}
```
DamienIrving: Introducing new things like `__file__` and R data packages when the reader has only just seen (possibly for the first time) how to create a python package seems like it might be confusing.

I think we need an exercise before `packaging-ex-data` that does nothing more than ask the reader to go through the basic process of creating a package from scratch. For example:

### Exercise \@ref(packaging-ex-pratchett) {-}

Each chapter in this book opens with a quote from British author Terry Pratchett. Here's a script `quotes.py` which contains a function `random_quote` which prints a random Pratchett quote.

```
import random

quote_list = ["""It’s still magic even if you know how it’s done.""",
              "Ninety percent of most magic merely consists "\
              "of knowing one extra fact.",
              """There isn’t a way things should be. """\
              """There’s just what happens, and what we do.""",
              "Multiple exclamation marks are a sure sign "\
              "of a diseased mind.",
              "+++ Divide By Cucumber Error. """\
              "Please Reinstall Universe And Reboot +++",
              """It’s got three keyboards and a hundred extra knobs, """\
              """including twelve with ‘?’ on them.""",
              "Evil begins when you begin to treat people as things.",
             ]
                         
def random_quote():
    """Print a random Pratchett quote."""
    print(random.choice(quote_list))
```
Create a new (conda) development environment called `pratchett` and `pip` install a new package called `pratchett` into that environment. The package should contain `quotes.py` such that once the package has been `pip` installed the user could run:

```
from pratchett import quotes

quotes.random_quote()
``` 

### Solution

The directory tree for the `pratchett` package is:
```
pratchett
├── pratchett
│   └── quotes.py
├── README.md
└── setup.py
```

The `README.md` file should contain a basic description of the package and how to install/use it, while `setup.py` should read as follows:
```
from setuptools import setup


setup(
    name='pratchett',
    version='0.1',
    author='Amira Khan',
    packages=['pratchett'],
)
```
The following sequence of commands is required to create the development environment, activate that environment and then pip install the package:
```
$ conda create -n pratchett python
$ conda activate pratchett
(pratchett)$ cd pratchett
(pratchett)$ pip install -e .
```

chapters/solutions.Rmd
```diff
@@ -1582,7 +1654,53 @@ most items would be checked off of the list.
...
+so it can look for other files in that directory.
+This means that we don't have assume the package has been installed in a particular place
+or that some environment variable will tell us where to look for our data.
+To test this idea out:
+
+1.  Create a directory called `data-package`.
+
+1.  Put a copy of `collated.csv` in that directory.
+
+1.  Add a file called `collated.py` to that directory.
```
DamienIrving: Is this confusing because we already have a script we've been working on called `collate.py`?

Maybe it would be better if this data package contained the word count csv files for a number of our books (e.g. Dracula, Jane Eyre, etc) rather than the collated count?

chapters/solutions.Rmd
```diff
@@ -1582,7 +1654,53 @@ most items would be checked off of the list.
...
 ### Exercise \@ref(packaging-ex-data) {-}
 
-FIXME
+The central idea is that a Python file can find out what directory it was read from,
+so it can look for other files in that directory.
+This means that we don't have assume the package has been installed in a particular place
+or that some environment variable will tell us where to look for our data.
+To test this idea out:
+
+1.  Create a directory called `data-package`.
```
DamienIrving: In the solution we probably need to describe the full directory tree for the new package:
```
data-package
├── data-package
|   ├── collated.csv
│   └── collated.py
├── README.md
└── setup.py
```
as well as the contents of `setup.py`:
```
from setuptools import setup


setup(
    name='data-package',
    version='0.1',
    author='Amira Khan',
    packages=['data-package'],
)
```

chapters/solutions.Rmd
```diff
@@ -1582,7 +1654,53 @@ most items would be checked off of the list.
...
+
+    ```python
+    import collated
+
+    data = collated.getData()
+    for row in data:
+        print(row)
+    ```
+
+1.  Run the test program.
```
DamienIrving: In the solution we probably need to describe the sequence of commands required to create a test environment and then pip install `data-package` into that environment:
```
$ conda create -n data python
$ conda activate data
(data)$ cd data-package
(data)$ pip install -e .
```

DamienIrving (changes requested):
Thanks, @gvwilson. I like the changes you've made to the config exercises and solutions - I've just made a few minor comments. (In answer to your question, we do want these changes in the main `zipf` directory. I've been updating the code in `zipf` anytime an exercise adds functionality to one of our scripts.)

I think the data package exercise is introducing too much new stuff at once. I'm not sure that we want/need that exercise at all, but if we do retain it I think we need a simpler exercise before it. In the comments I've written out a new exercise/solution for a `pratchett` package - feel free to add it in if you like it.

gvwilson:
@DamienIrving thanks for the feedback - I removed the data package example entirely and substituted the pratchett quotes, and made other changes as suggested.

DamienIrving (approved):
Looks good, @gvwilson. Ready to merge :smile:

PR #502, opened 2020-10-01
Adding/editing testing exercises

DamienIrving:


chapters/solutions.Rmd
```diff
@@ -1430,6 +1430,95 @@ def test_fails_for_non_list():
 
 If the call to `total` *doesn't* raise `AssertionError`, this test fails.
 
+### Exercise \@ref(testing-ex-error-handling) {-}
+
+1. `collate.py` should read as follows:
```
gvwilson: Looking at the book in print, we're repeating all of our scripts with slight variations several times - in cases like this, can we just show the bits that have changed? Or will that confuse?

chapters/testing.Rmd
```diff
@@ -1165,6 +1166,69 @@ def total(values):
 Write unit tests to check that the `assert` statements in the previous question's `total` function
 do what they're supposed to.
 
+### Test error handling {#testing-ex-error-handling}
+
+In Chapter \@ref(errors), we added some functionality to `collate.py` in order to handle
```
gvwilson: "we modified `collate.py` to handle" rather than "we added some functionality to `collate.py` in order to handle"

chapters/testing.Rmd
```diff
@@ -1165,6 +1166,69 @@ def total(values):
...
+different types of errors associated with reading input files.
+The relevant code appears in the `main` function of that script:
```
gvwilson: "in `main`" instead of "in the `main` function of that script"

chapters/testing.Rmd
```diff
@@ -1165,6 +1166,69 @@ def total(values):
...
+        except FileNotFoundError:
+            logging.warning(f'{file_name} not processed: File does not exist')
+        except PermissionError:
+            logging.warning(f'{file_name} not processed: No permission to read file')
+        except Exception as error:
+            logging.warning(f'{file_name} not processed: {error}')
+    utilities.collection_to_csv(word_counts, num=args.num)
+```
+
+In this current chapter we discussed using `pytest.raises`
```
gvwilson: "this chapter" rather than "this current chapter"

chapters/testing.Rmd
```diff
@@ -1165,6 +1166,69 @@ def total(values):
...
+to test whether the error handling in a program is working as expected (Section \@ref(testing-failure)).
+An appropriate test for `collate.py` would be to check that an `OSError` is raised
+when the input file does not end in `.csv`.
+The problem is that it is difficult to write a simple unit test
+for the code as it is currently written,
+because `main` is a long function that requires command line arguments as input.
+In other words, we need to edit `collate.py` so that it is more modular
```
gvwilson: I would move this sentence to the opening of instruction 1 below.

chapters/solutions.Rmd
DamienIrving: Good point. I think in general we need to give the full version of the script in the question/exercise (i.e. we can't expect the reader to look back and figure out what the script should look like), and then in the solutions only provide information about the bits that have changed. I've now done that in this PR and I'll create an issue to make sure we are being consistent with exercises and solutions in other chapters.

DamienIrving:
Ready for review.

PR #513, opened 2020-11-11
Splitting bash-basics in two

### LINKED ISSUE #507: Split the "Basics of the Unix Shell" chapter in two
The Basics of the Unix Shell (Chapter 2) is too long. It needs to be separated into two chapters, splitting between Sections 2.8 and 2.9. The first chapter will cover single bash commands and the second combining multiple bash commands.

gvwilson:
Closes #507.


lwjohnst86 (approved):
Looks good!

PR #517, opened 2020-11-11
Removing contributor guide from book

gvwilson:
Note: this will _also_ remove the contributor's guide from the web version, which we may not want, but trying to conditionally include a section in the HTML without messing up the appendix numbering (or lettering) in the PDF is beyond me.


lwjohnst86:
I think its fine that it isn't on the web version, as long as there is a link to it somewhere (which we will probably add in the Welcome section).

lwjohnst86 (approved):

PR #518, opened 2020-11-11
Second review

gvwilson:


reviews/review-02+2020-11-11.md
```diff
@@ -0,0 +1,77 @@
...
+## What reservations do you have about this manuscript?
+
+There is a sense in which the intended audience may not view
+themselves as software engineers (i.e., they are likely researchers in
+some other specific scientific discipline). As such, the term
+"research software engineering" may miss the mark. That said, there
+are plenty of people who do use this term and there are not really any
+effective alternatives in common use. So the title is still likely to
+reach it's intended audience.
+
```
cwickham, line 18: I think the Introduction does a good job of describing what a "research software engineer" is, so a reader doesn't have to get very far to figure out if this is for them, but I still think this is an interesting point.  Are we missing out on people who aren't familiar with the term "research software engineering", or don't recognize what they do (or want to do) is "research software engineering"?

Is it worth giving our title one last chunk of thought?  I'm not a huge fan of colons, but maybe a subtitle might help.  Stealing straight from the Intro: "Building software that makes research possible"

DamienIrving: I'm also a little worried about our use of the term "research software engineer," but I'm also not sure that there is a better alternative. 

From my limited participation in eResearch conferences and national RSE associations, there seems to be two paths by which people come into the research software space:
1. Researchers who pick up the software skills later on
2. Software engineers who pick up the research knowledge later on

The latter would definitely call themselves "Research Software Engineers" and in my experience they tend to dominate the active membership of national RSE associations. Our book isn't for these people. Our book is for researchers who are trying to pick up software skills, who are less likely to identify with the RSE label. Perhaps we need to make this distinction clearer in the book prologue?

An alternative book title might do away with the word engineering and just be: 
"Writing Research Software with Python"

reviews/review-02+2020-11-11.md
```diff
@@ -0,0 +1,77 @@
...
+I'm impressed at the depth and breadth of the content, so I see no
+need for major revisions beyond normal copy-editing, finishing the
+figures, etc.
+
+## Are the chapter titles/subtitles appropriate?
+
+The chapter titles form an appropriate sequence. I might consider
+swapping Chapters 12 (Provenance) and 13 (Python Packaging), but the
+remaining topics build in a sensible order.
+
```
DamienIrving, line 30: My preference would be to leave the order of the Provenance and Packaging chapters as is. The main reason for this is that if we do provenance after packaging, it kind of implies that the code associated with a research paper can't be made available unless it is packaged for a general audience, which is definitely not the signal we want to send.

reviews/review-02+2020-11-11.md
gvwilson: I'd like to keep RSE in the title - the term has some traction, and while people who already have these skills may dominate the associations, I think there's a much larger (quieter) group who are trying to acquire them and become RSEs. I share @cwickham's dislike of colons, but I like the subtitle idea a lot.

lwjohnst86 (commented):
I don't have any other comments from what the others have written. I also like the subtitle option suggested by @cwickham.

reviews/review-02+2020-11-11.md
gvwilson: I have no strong opinion - if no one else strongly prefers a change, let's leave it as is.

PR #525, opened 2020-11-11
Correcting sample filesystem figure

### LINKED ISSUE #500: Bug in filesystem figure
In [`figures/bash-basics/sample-filesystem.png`](https://github.com/merely-useful/py-rse/blob/book/figures/bash-basics/sample-filesystem.png) there's a `/Users/tmp/` directory, but later on when the following command is run there's no `tmp` directory.

```
$ ls /Users
amira   jun     sami
```

k8hertweck:
From unix basics chapter, resolves #500


DamienIrving (approved):
👍

PR #531, opened 2020-11-12
Fix long code chunks

lwjohnst86:
Resolves #469

PR #533, opened 2020-11-15
Line-wrapping verbatim text

gvwilson:


lwjohnst86:
@gvwilson I already did this in #531 (since the verbatim text was in code chunks). Closing.

PR #538, opened 2020-11-16
Use shaded box for callout box to be prettier.

### LINKED ISSUE #521: Make callout boxes look better in PDF version
@gvwilson will chase this up with the publisher when we're done other changes.

lwjohnst86:
What do you all think of this callout box? Instead of the horizontal line above and below the text.

Screenshot from 2020-11-16 14-40-32

Resolves #521


gvwilson:
:+1: I like it a lot.

PR #539, opened 2020-11-16
Reworking Unix figures

k8hertweck:
A few adjustments to Unix diagrams, so they'll coordinate better with the Git figures I'm preparing.I also included draw.io files and all export options as directed in the contribution instructions. There are a few others involving screenshots I've left as only png since draw.io doesn't seem to play well with cropping and resizing images.


gvwilson:
The pipe figure is beautiful.

PR #541, opened 2020-11-24
More work on long code blocks

DamienIrving:
Following on from #531, this PR makes sure our scripts are PEP-8 compliant and fixes some more code blocks that are too long.


DamienIrving:
@gvwilson Do you think it would be possible to get a hard character limit from the publisher for lines in code blocks? Without a hard limit it's difficult to be confident that we won't have to come back do more line shortening later, which is a problem because changes to our scripts can have flow on effects on `git diff` output (for example) in subsequent chapters.

gvwilson:
@DamienIrving I believe the krantz.cls style file we're using is the one we will use in production, so if the lines wrap correctly when we build, we're good. I will double check w/ Randi Cohen.

gvwilson:
```
Hi Greg,

Yes, we can definitely use the Krantz style for the book. It is one of our common styles, so it is fine. No need to re-wrap anything!

Randi
```

DamienIrving:
Ok. This one is ready for review.

It makes sure that: 
* Each line in our input code blocks (whether they be `python`, `bash`, `diff`, `markdown`, `makefile` or `yaml`) are less than 66 characters, which is the limit for the latex style we are using
* If the output from our code blocks (`text` blocks) goes over 66 characters, it's only a little over and doesn't go off the edge of the page 
* Our scripts pass the `pycodestyle` test

The reason I'm ok with the output blocks occasionally going a little over 66 characters is that those blocks don't have a grey shaded background, which means it's not so obvious that they've gone over (and changing them to be < 66 would look very weird in many cases).

gvwilson:
Heroic work - :+1:.

gvwilson (approved):
Awesome work - thank you.

PR #545, opened 2020-11-24
Defining some Unicode box-drawing characters for tree output.

### LINKED ISSUE #543: Difference in directory tree between HTML and PDF. Fix.

gvwilson:
Closes #543.
(Almost - some of the vertical bars don't render cleanly even though they seem to be the same character.)


DamienIrving (commented):
So should we merge this PR and then try and figure out what's up with the vertical bars that still won't render properly?

DamienIrving (approved):
Sorry, my previous comment was meant to be an "approve" not "comment"

gvwilson:
:+1:

PR #548, opened 2020-11-25
Introduce error handling at a high level up front

joelostblom:
@ChristinaLK Please have a look and let me know if this summary of the sections if helpful for building a mental model of error handling before digging into each section's specifics.

close #512

PR #550, opened 2020-11-26
Error overview revised

gvwilson:
#548 added a short introduction to error handling to the top of the chapter, but the content duplicated material from further down. This PR moves that existing material to the top of the chapter (including glossary links and index terms) and turns the bit that was left over into a callout box.

Replaces #548.


DamienIrving (commented):
The new introduction to the error handling chapter makes multiple forward references to the testing chapter. Perhaps that's a sign that the testing chapter should come before the error handling chapter? 

If we changed the order, the introduction to the error handling chapter would explain that in the previous testing chapter we were mainly dealing with internal errors. When we found an error, we'd halt the program (typically using assertions). In this chapter, we will explore the options for handling errors. Rather than simply halt the program when an error occurs, there are occasions where we'd like our program to continue (perhaps with altered behavior) when certain types of errors arise. Error handling (as opposed to error halting) is more often associated with external errors...

Switching the order of the error handling and testing chapters would be relatively trivial. The only thing that would need to be changed is that the short Section 12.5 (Testing Error Handling) and the associated Exercise 12.12.3 would need to be moved from the testing chapter to the error handling chapter.  

(The other bonus is that the error handling chapter is currently pretty short and the testing chapter pretty long, so moving a bit of content from testing to error handling would be a good thing.)

k8hertweck:
I wouldn't mind this change; it makes sense logically to talk internal errors first, then broaden out to external.

I'm working on Amira's repo tomorrow, so I'll check back to see if we've decided to make this update so I'll know how to proceed.

gvwilson:
This PR swaps the order of the chapters and moves one exercise to match. There is no longer any mention of `exception` in the testing chapter, and no forward references from it to the errors chapter. Please let me know if I've missed anything.

chapters/testing.Rmd
```diff
@@ -1095,7 +1036,7 @@ TDD's advocates claim that this leads to better code because:
 
 These arguments are plausible.
 However,
-studies such as @Fucc2016 and @Fucc2017 don't support them:
+studies such as [@Fucc2016; @Fucc2017] don't support them:
```
DamienIrving: I think the original formatting was correct:
"studies such as Fucci et al. (2016) and Fucci et al. (2017) don’t..."

You've changed it to:
"studies such as (Fucci et al. 2016, 2017) don’t..."

chapters/errors.Rmd
```diff
@@ -308,54 +345,82 @@ What we *should* do is check that there are two columns separated by a comma,
...
+> that the program doesn't have permission to write to an output directory.
+> It's a little extra work to check things like this up front,
+> but the larger your program or the longer it runs,
+> the more useful those checks will be.
+
+## Testing Error Handling {#errors-testing}
+
+An alarm isn't much use if it doesn't go off when it's supposed to.
+Equally,
+if a function doesn't raise an exception when it should (Section \@ref(errors-exceptions)),
```
DamienIrving:
```suggestion
if a function doesn't raise an exception when it should,
```
We don't need the Section reference anymore because this content is now in the error handling chapter.

chapters/errors.Rmd
```diff
@@ -308,54 +345,82 @@ What we *should* do is check that there are two columns separated by a comma,
...
+errors can easily slip past us.
+If we want to check that a function called `func`
+raises an `ExpectedError` exception\index{testing!error handling}\index{exception!testing}
+we can use the following template:
```
DamienIrving:
```suggestion
we can use the following template for a unit test:
```

chapters/errors.Rmd
```diff
@@ -308,54 +345,82 @@ What we *should* do is check that there are two columns separated by a comma,
...
+>
+> The one rule we should *always* follow is to check for errors as early as possible
+> so that we don't waste the user's time.
+> Few things are as frustrating as being told at the end of an hour-long calculation
+> that the program doesn't have permission to write to an output directory.
+> It's a little extra work to check things like this up front,
+> but the larger your program or the longer it runs,
+> the more useful those checks will be.
+
+## Testing Error Handling {#errors-testing}
```
DamienIrving: Maybe the `#errors-testing` section should go after the `#errors-messages` section? I say that because the error messages section flows quite nicely from the `#errors-exceptions` section (i.e. we've just written our own error message `OSError(f'{fname} is not a CSV file.')` and now need to consider how to write good error messages).

So my suggested order for the chapter is:
1. `#errors-exceptions`
2. `#errors-messages`
3. `#errors-testing`
4. `#errors-logging`

chapters/errors.Rmd
```diff
@@ -815,6 +881,111 @@ the default name of `collate.log`.
 2.  Modify your first solution to handle nonexistent files
     and permission problems separately.
 
+### Testing error handling {#errors-ex-error-handling}
+
+In Chapter \@ref(errors), we modified `collate.py` to handle
+different types of errors associated with reading input files.
+The relevant code appears in `main`:
```
DamienIrving: Now that we've moved this exercise from the testing chapter into the error handling chapter, the modification to `collate.py` that is referred to here happens in the immediate previous exercise (`#errors-ex-exceptions`). 

We therefore probably just need to say something like:

In our suggested solution to the previous exercise, the `main()` function in `collate.py` reads as follows:

```
def main(args):
    """Run the command line program."""
    log_lev = logging.DEBUG if args.verbose else logging.WARNING
    logging.basicConfig(level=log_lev, filename=args.logfile)
    word_counts = Counter()
    logging.info('Processing files...')
    for fname in args.infiles:
        try:
            logging.debug(f'Reading in {fname}...')
            if fname[-4:] != '.csv':
                msg = util.ERRORS['not_csv_suffix'].format(
                      fname=fname)
                raise OSError(msg)
            with open(fname, 'r') as reader:
                logging.debug('Computing word counts...')
                update_counts(reader, word_counts)
        except FileNotFoundError:
            msg = f'{fname} not processed: File does not exist'
            logging.warning(msg)
        except PermissionError:
            msg = f'{fname} not processed: No permission to read'
            logging.warning(msg)
        except Exception as error:
            msg = f'{fname} not processed: {error}'
            logging.warning(msg)
    util.collection_to_csv(word_counts, num=args.num) 
```

chapters/errors.Rmd
```diff
@@ -815,6 +881,111 @@ the default name of `collate.log`.
...
+                        action="store_true", default=False,
+                        help="Set logging level to DEBUG")
+    parser.add_argument('-l', '--logfile',
+                        type=str, default='collate.log',
+                        help='Name of the log file')
+    args = parser.parse_args()
+    main(args)
+```
+
+In this chapter we discussed using `pytest.raises`
```
DamienIrving:
```suggestion

```

chapters/errors.Rmd
```diff
@@ -815,6 +881,111 @@ the default name of `collate.log`.
...
+to test whether the error handling in a program is working as expected (Section \@ref(errors-testing)).
```
DamienIrving:
```suggestion

```

chapters/errors.Rmd
```diff
@@ -815,6 +881,111 @@ the default name of `collate.log`.
...
+2.  Add a unit test to `test_zipfs.py` that uses `pytest.raises` to check that
+    the new `collate.process_file` function raises an `OSError` if the input
+    file does not end in `.csv`.  Run `pytest` to check that the new test
+    passes.
+
+3.  Add a unit test to `test_zipfs.py` that uses `pytest.raises` to check that
+    the new `collate.process_file` function raises a `FileNotFoundError` if the
+    input file does not exist.  Run `pytest` to check that the new test passes.
+
+4.  Use the `coverage` library to check that the relevant commands in
```
DamienIrving:
```suggestion
4.  Use the `coverage` library (Section \@ref(testing-coverage)) to check that the relevant commands in
```

DamienIrving (approved):
Thanks, @gvwilson. I've made a few minor comments, but otherwise I like this switch in chapter order. I think it helps people understand error handling if they've seen the testing content first (i.e. they've done some basic "error halting" with assertions before learning about "error handling"). The order that the solutions appear in `solutions.Rmd` will also need to changed to reflect the fact that the testing chapter now comes before error handling, but I think that can be tackled as part of #555.

chapters/testing.Rmd
gvwilson: :+1:

chapters/errors.Rmd
gvwilson: :+1:

chapters/errors.Rmd
gvwilson: :+1:

chapters/errors.Rmd
gvwilson: :+1:

chapters/errors.Rmd
gvwilson: :+1:

chapters/errors.Rmd
gvwilson: I think it reads better if I take out this whole para.

chapters/errors.Rmd
gvwilson: :+1:

gvwilson:
All changes made, and the order of the sections in `solutions` has been swapped (thanks for catching that).

PR #558, opened 2020-11-29
Appendix references and order

DamienIrving:
A number of the Appendices weren't referred to at all in text, so I've added references where needed.

I've also updated the order of the appendices so they appear in the order that they are first referenced in text.


_bookdown.yml
```diff
@@ -28,15 +28,15 @@ rmd_files:
 
   # End files
   - appendix.Rmd
-  - glossary/glossary.md
+  - chapters/solutions.Rmd
   - chapters/objectives.Rmd
   - chapters/keypoints.Rmd
-  - chapters/solutions.Rmd
-  - chapters/anaconda.Rmd
   - chapters/tree.Rmd
```
k8hertweck, line 34: I think style comes before tree?

DamienIrving: There was a random/unnecessary reference to style in the project structure section, so I took it out and added a reference in the scripting chapter instead, which is where we first write some Python code and is thus the most appropriate place for the initial reference to that Appendix to be.

k8hertweck: Got it, I should really stop trying to review when I'm in the middle of a hairy PR of my own.

k8hertweck (approved):
Thanks!

PR #561, opened 2020-11-30
Prefix code blocks with ':' not ','

gvwilson:
There were a few places where we ended a sentence introducing a code block with ',' instead of ':' - this makes them consistent.

### DIFF
--- a/chapters/config.Rmd
+++ b/chapters/config.Rmd
@@ -233,7 +233,7 @@ mpl.get_configdir()
 /Users/amira/.matplotlib
 ```
 
-Once we've created the new sub-directory,
+Once we've created the new sub-directory:
 
 ```bash
 $ mkdir /Users/amira/.matplotlib/stylelib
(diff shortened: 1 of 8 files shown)


chapters/config.Rmd
```diff
@@ -233,7 +233,7 @@ mpl.get_configdir()
 /Users/amira/.matplotlib
 ```
 
-Once we've created the new sub-directory,
+Once we've created the new sub-directory:
```
DamienIrving, line 236: I think it's pretty uncontroversial to say that we should use `:` when the code block comes at the end of a sentence, but doesn't it make more sense to use `,` when the code block is introduced mid-sentence, like it is in this example?

DamienIrving (commented):
Thanks for raising this issue - I've made one comment, which pretty much sums up my thoughts :-)

chapters/config.Rmd
gvwilson: This. Is. PYTHON!!

DamienIrving: 🤣

PR #560, opened 2020-11-30
Revisions following test run

k8hertweck:
Starting with git-cmdline, when Amira's repo is tracked, these are revisions that reconcile what the final repo will look like, the scripts in this repo's `zipf` directory, and exercises/solutions. It's a continuation of #554 and continues resolving issues mentioned there.

There were more changes than I anticipated. Major alternations include:

- updating git output messages and commit numbers
- updating exercises to avoid readers having conflicts in future chapters
- updating solutions to match current exercises

I've been thinking it might be useful to add [lightweight tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging) in [Amira's repo](https://github.com/amira-khan/zipf) to label the code in different chapters, and to differentiate the "hard copy" version of code from books.

PR #569, opened 2020-12-09
Update packaging chapter to avoid package name clash

### LINKED ISSUE #410: Discussion: How should we import our utilities module?
In the packaging chapter we currently say/do the following:

"For our scripts to work as a Python package, we only need to make one important change to the code itself: changing the syntax for how we import `utilities`. Currently, both `collate.py` and `countwords.py` contains this line:

```python
import utilities
```

This is called an implicit relative import, because it is not clear whether there exists a Python package called `utilities` or there is a file in our local directory named `utilities.py` (as in our case). This ambiguity is not ideal and when making a package we need to be more explicit and write:

```python
from zipf import utilities
```
"
...

The problem is that Read the Docs hates this and thus won't display `collate` and `countwords` in the module index:
```
WARNING: autodoc: failed to import module 'collate'; the following exception was raised: No module named 'zipf'
WARNING: autodoc: failed to import module 'countwords'; the following exception was raised: No module named 'zipf'
```

What should we do about this?

### LINKED ISSUE #568: Should we publish our zipf package to PyPI and Read The Docs?
In the packaging chapter we talk about uploading our `zipf` package to PyPI and documentation to Read The Docs, but we don't actually do it. I'm wondering if that's how we want to leave things, or if we do want to actually have it out there as a package (linked to Amira's GitHub account)?

One problem is that there's already a package called `zipf`, so we'd have to pick a different name like `zipfpy` or `pyzipf`.
https://pypi.org/project/zipf/
https://zipf.readthedocs.io

I guess the advantage of having our package out there is that if Amira doesn't publish it to PyPI and Read The Docs, someone reading our book will no doubt do it. In other words, even if we include call out boxes telling people following along that they shouldn't actually hit the red button and publish to PyPI and Read The Docs, someone will inevitably hit the red button.

What do people think? I hope I've raised this issue before @k8hertweck gets to the packaging chapter in #560...

DamienIrving:
Resolves #568
Resolves #410

As part of #560, Amira will need to create the following pages:
https://test.pypi.org/project/pyzipf
https://pypi.org/project/pyzipf/
https://pyzipf.readthedocs.io/

That will avoid the problem of random people creating those pages when following the commands in the book.

(@k8hertweck - let me know if you have any issues with the Read The Docs build and I can help out.)

### DIFF
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -637,6 +637,9 @@ and will vary from computer to computer.
 > but the general sequence of commands in this section
 > is an excellent resource to refer back to when you are
 > uploading your own packages.
+> If you want to try uploading your own package via `twine`,
+> you could suffix it with your name: `pyzipf-yourname`
+> and use the TestPyPI repository for the upload.
 
 An installable package is most useful
 if we distribute it so that anyone who wants it
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -606,7 +606,7 @@ import sys
 sys.path
 ```
 
-```
+```text
 ['',
 '/Users/amira/opt/anaconda3/envs/pyzipf/lib/python37.zip',
 '/Users/amira/opt/anaconda3/envs/pyzipf/lib/python3.7',
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -123,7 +123,7 @@ $ mv bin pyzipf
 >
 > ```bash
 > $ git remote set-url origin 
-    https://github.com/amira-khan/pyzipf.git  
+>   https://github.com/amira-khan/pyzipf.git  
 > ```
 
 Python has several ways to build an installable package.\index{Python package!building}
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -240,6 +240,16 @@ such as `bin` and `lib`.
 It also creates `~/opt/anaconda3/envs/pyzipf/bin/python`,
 which checks for packages in these directories before checking the main installation.
 
+>**Anaconda location variations**
+>
+> The default location for Anaconda differs between operating systems.
+> Our examples show the default location on MacOS (`/Users/amira/opt/anaconda3`),
+> but on Linux it is typically `/home/amira/anaconda3` and
+> on Windows it might be `C:\Documents and Settings\amira\anaconda3`
+> or `C:\Users\amira\anaconda3` (depending on the version of Windows).
+> During the installation process users can also choose a custom location
+> if they like (Section \@ref(getting-started-install-software)).
+
 We can switch to the `pyzipf` environment by running:\index{virtual environment (in Python)!switching}
 
 ```bash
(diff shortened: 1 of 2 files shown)


DamienIrving:
This one is ready for review.

chapters/packaging.Rmd
```diff
@@ -315,37 +289,37 @@ Let's install our package in this virtual environment.\index{Python package!inst
...
 ```
 
 The `-e` option indicates that we want to install the package in "editable" mode,
 which means that any changes we make in the package code are directly available to use
 without having to reinstall the package;
 the `.` means "install from the current directory".
 
-If we look in `~/opt/anaconda3/envs/zipf/lib/python3.8/site-packages/`,
-we can see the `zipf` package beside all the other locally-installed packages.
+If we look in `~/opt/anaconda3/envs/pyzipf/lib/python3.8/site-packages/`,
```
joelostblom, line 331: Just a note that anaconda is in `opt` only for mac users. It would be in `~` directly on linux and in `C:\Users\Username` I believe on Windows. I don't know if we do/should mention this somewhere in the beginning what operating system we are on, I didn't see anything when searching for it. I think we currently have some differences between chapters, but not sure if they are noticeable enough to be confusing (e.g. I saw that the scripting package uses python on linux), probably not.

chapters/packaging.Rmd
```diff
@@ -648,63 +622,79 @@ and will vary from computer to computer.
...
+> <https://test.pypi.org/project/pyzipf>
+>
+> <https://pypi.org/project/pyzipf/>
+>
+> You won't be able to execute the `twine upload`
+> commands exactly as shown
+> (because Amira has already uploaded the `pyzipf` package),
+> but the general sequence of commands in this section
+> is an excellent resource to refer back to when you are
+> uploading your own packages.
```
joelostblom, line 649:
```suggestion
> uploading your own packages.
> If you want to try uploading your own package via `twine`,
> you could suffix it with your name: `pyzipf-yourname`
> and use the TestPyPI repository for the upload.
```

joelostblom (approved):
Minor comments, looks good to me as is also!

chapters/packaging.Rmd
DamienIrving: In Chapter 2 we have a call-out box to explain why we are using `/Users/amira` as our home directory:

> **Home Directory Variations**
>
> Our home directory will be in different places on different operating systems. On Linux it may be `/home/amira`, and on Windows it may be `C:\Documents and Settings\amira` or `C:\Users\amira` (depending on the version of Windows). Our examples show what we would see on MacOS.

We could have one in this chapter that reads as follows?

>**Anaconda location variations**
>
> The default location for Anaconda differs between operating systems. Our examples show the default location on MacOS (`/Users/amira/opt/anaconda3`), but on Linux it is typically `/home/amira/anaconda3` and on Windows it might be `C:\Documents and Settings\amira\anaconda3` or `C:\Users\amira\anaconda3` (depending on the version of Windows). During the installation process users can also choose a custom location (Section 1.3).

chapters/packaging.Rmd
DamienIrving: I wasn't sure whether to encourage people to clog up TestPyPI and PyPI with their own copy of the zipf package or not? I guess it doesn't really matter if they do. We could encourage them to delete their copy when they're done trying it out?

chapters/packaging.Rmd
joelostblom: Yes that looks great!

chapters/packaging.Rmd
joelostblom: I had the same thought, maybe recommend it for test-pypi specifically? I updated my suggestion.

joelostblom: TestPyPI should be cleaned every now and then, so I don't think they need to delete anything https://packaging.python.org/guides/using-testpypi/#registering-your-account

DamienIrving:
@k8hertweck - the packaging chapter is now ready for your run though in #560 😄

PR #573, opened 2020-12-13
Appendix proof-read

DamienIrving:


book.bib
```diff
@@ -144,7 +134,7 @@ @book{Bran1995
 @misc{Broc2019,
   author = {Jon Brock},
   year = {2019},
-  title = {"A love letter to your future self": What scientists need to know about FAIR data},
+  title = {A Love Letter to Your Future Self: What Scientists Need to Know About {FAIR} Data},
```
gvwilson: I think the quotation marks are in the title of the original?

chapters/anaconda.Rmd
```diff
@@ -2,7 +2,7 @@
 
 When people first started using Python for data science,
 installing the relevant libraries could be difficult.\index{package installation}
-The main problem was that the Python package installer ([pip][pip])\index{pip}
+The main problem was that the Python package installer ([`pip`][pip])\index{pip}
```
gvwilson, line 5: I approve the change, but the word nerd in me can't help but notice that we use Git, Python, and Make rather than `git`, `python`, and `make` when referring to the tools. Ah, English...

chapters/documentation.Rmd
```diff
@@ -11,6 +11,59 @@ the best function names in the world aren't going to answer the questions
...
+That one- or two-liner should begin with an active verb and describe either
+how inputs are turned into outputs,
+or what side effects the function has;
+as we discuss below,
+if we need to describe both,
+we should probably rewrite our function.
+
+An active verb is something like "extract", "normalize", or "find".
+For example,
+these are all good one-line docstrings:
```
gvwilson, line 40: I like this section a lot - is it worth adding a couple of examples of *bad* docstrings for contrast?

gvwilson (approved):
Great work - thank you.

book.bib
DamienIrving: Yeah, I wasn't sure what to do here. There are quotation marks in the original title, but our referencing style in the PDF puts quotation marks around the title of articles, so it ends up formatting with an open and closed quotation at the beginning (i.e. ""A love...) and then an open quotation after self and a closed at the end...

chapters/anaconda.Rmd
DamienIrving: I had exactly the same thought. I think (with the few edits I've made in this PR) we've been consistent with which things are `software` as opposed to Software (i.e. there shouldn't be a mix of conda and `conda` anymore), but I agree that technically Git, Python and Make should be `git`, `python` and `make`. I guess what we've essentially done is use `software` in all cases except for things that we mention frequently in text.

chapters/documentation.Rmd
DamienIrving: For each of the three good examples, perhaps we should show what a bad docstring for the same function would be?

The other glaring omission from this section is that we give a five point list of things to include in docstrings (name and purpose of every parameter, etc), but we don't give an example of what a docstring with all that information in it looks like, nor do we talk about the option of using reStructuredText, Google or Numpydoc style (e.g. [this post](http://daouzli.com/blog/docstring.html)) to format that information.

DamienIrving:
@lwjohnst86 Is there a way to include some spacing between each of the references in the online reference list (and/or indentation for the second line of a reference onwards)? At the moment it's all a bit cramped. 
Screen Shot 2020-12-14 at 7 44 11 am

book.bib
lwjohnst86: What about using single quotes?

lwjohnst86:
@DamienIrving this is handled I believe by the CSL format. I changed format to chicago which makes more space between. See #574

book.bib
gvwilson: I believe the double double-quotes are canonical, but yeah, it looks weird, so let's go with single quotes just to give future bibliographic analysis software a heart attack.

chapters/documentation.Rmd
DamienIrving: I might merge this PR as is and start a new for the suggested docstring content...

DamienIrving:
In the glossary, many entries reference other entries. These keywords are clearly highlighted in the html version of the book (because they are hyperlink blue), but aren't highlighted at all in the pdf version. Is that a problem? 

HTML:
Screen Shot 2020-12-17 at 5 03 35 pm

PDF:
Screen Shot 2020-12-17 at 5 03 49 pm

gvwilson:
I'll see if I can find a quick fix - otherwise, I vote we ship as-is.

Later: fixed in https://github.com/merely-useful/py-rse/pull/575

gvwilson:
No quick fix - I vote we merge and ship.

PR #574, opened 2020-12-14
use IEEE style

lwjohnst86:
To deal with comment from #573


DamienIrving (changes requested):
The reference list looks better in the html version now (there's nice spacing between each entry), but in the pdf version it now does this weird thing where it gives a footnote URL for each DOI, rather than just listing the DOI in the reference itself prefixed by `https://doi.org/` (which is what the current referencing in the `book` branch does).

New pdf reference list:
Screen Shot 2020-12-15 at 7 31 39 am

Old pdf reference list:
Screen Shot 2020-12-15 at 7 39 54 am


Also, this PR includes a very old version of `index.Rmd`, so it reverses a whole bunch of improvements we've made to that file over the past few weeks/months (e.g. it removes the book subtitle, adds back in a whole bunch of content that was moved to the "Getting Started" chapter, etc).

lwjohnst86:
Oh damn. I even pulled from latest... weird, I'll look into that. I'll look more into the citation format as well.

gvwilson:
Any updates on this one?

lwjohnst86:
@gvwilson still looking into it. Its a tricky one.

lwjohnst86:
Fixed! I changed the style to IEEE, but its pretty similar to chicago. It isn't indented tho. Here's what it looks like now:

Screenshot from 2020-12-19 12-31-43

DamienIrving (commented):
That spacing looks nice in the html version.

I couldn't check the pdf version due to #578

PR #581, opened 2020-12-21
Resolving errant comments

k8hertweck:
I missed a few comments from the review of #560 . This resolves them, and I've marked resolved in the other thread as well.


DamienIrving (approved):
👍

DamienIrving:
@k8hertweck I merged before realizing that `.travis.yml` in the testing chapter still specifies Python 3.6. Updating it to 3.7 would require the corresponding figure to be updated as well so if updating it is a hassle we could let it slide, but just thought I'd flag it in case it's not too much trouble.

PR #583, opened 2020-12-21
Travis python version update

k8hertweck:
Following @DamienIrving 's [comment](https://github.com/merely-useful/py-rse/pull/581#issuecomment-748876276), we needed to address my mistake in not updating the python version passed to Travis CI. As per @gvwilson 's suggestion, added a pop-out that describes how this happened and makes it ok.


lwjohnst86 (commented):
I like this, it gives a good view of that in real life, things change, especially with software. :+1:

PR #585, opened 2020-12-22
Fixing some LaTeX issues.

gvwilson:
1.  Including `footnote.sty` so that `\savenotes` will work.
2.  Updating `krantz.cls` so that links in ToC jump to correct place (rather than one past correct place).


lwjohnst86 (approved):

lwjohnst86:
Nice you got the TOC links to work correctly, great work!

PR #591, opened 2021-01-10
Fix image size problems

lwjohnst86:
Figures 2.2, 2.5, 3.1, 3.2, and E.1 had their pdf images with extra space around the edges (from the export from draw.io). Fixed it so only the image was exported.


DamienIrving (commented):
Nice fix.

Does Figure 11.3 also need to be made larger?

gvwilson:
Just rebuilt - the changes all look good to me, but I think the screenshot of Nano and the structure of a man page will both be clearer if they are dark-on-white instead of white-on-dark - can you please recapture and stuff 'em into this PR and I'll re-review?

lwjohnst86:
Regarding, the dark on white, the man page has extra stuff added to it and also looks different from mine when I do `man wc`. @k8hertweck you did this image, do you mind switching the man background to dark on white?

k8hertweck:
Yup, I can do this

gvwilson:
Thanks - gi's a shout when it's in?

k8hertweck:
@gvwilson @lwjohnst86 Added the man image; let me know if it needs to be tweaked (I didn't check in the rendered book).

gvwilson:
Thanks very much - the only one that's still white-on-black is the first one (Fig 2.1, "The Bash shell"). Can someone please re-do that and then merge this? Everything builds fine for me with the new figs.

k8hertweck:
I've got it.

k8hertweck:
@gvwilson Done!

PR #592, opened 2021-01-11
Fix overflow lines

DamienIrving:
I've fixed the places identified in the pre-production review where the text runs over the margin.

I've been having trouble building the pdf on my computer for a few weeks now, so when this is merged we just need to check that changing `howpublished` to `url` in all the `@misc` references in `book.bib` has indeed fixed the overflow issues with URLs in the reference list of the pdf document.


gvwilson:
Built the PDF locally and checked the bibliography and a few other fixes - all looks good. Also fixed one broken `\index` entry (my fault).

PR #593, opened 2021-01-11
Fixing typos

sida-wang:
Correcting a few typos that I came across while going through the first 11 chapters. 

The change in bash-tools.Rmd might require further adjustment to the lines following that. Default sort when I ran the code was alphabetical so the only way to replicate the results was to use the -n option.

The change in teams.Rmd can also be 'In typical research software projects,', I just went with one of the two options, I can change this based on feedback. 

Remaining changes should be self-explanatory but happy to discuss if any don't make sense.

### DIFF
--- a/chapters/bash-tools.Rmd
+++ b/chapters/bash-tools.Rmd
@@ -90,7 +90,7 @@ $ cat lengths.txt
 We can now use `sort` to sort the lines in this file:
 
 ```bash
-$ sort lengths.txt
+$ sort lengths.txt -n
 ```
 
 ```text
--- a/chapters/bash-tools.Rmd
+++ b/chapters/bash-tools.Rmd
@@ -106,7 +106,7 @@ $ sort lengths.txt -n
 ```
 
 Just to be safe,
-we should use `sort`'s `-n` option to specify that we want to sort numerically.
+we use `sort`'s `-n` option to specify that we want to sort numerically.
 Without it,
 `sort` would order things alphabetically
 so that `10` would come before `2`.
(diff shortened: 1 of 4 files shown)


chapters/bash-tools.Rmd
```diff
@@ -90,7 +90,7 @@ $ cat lengths.txt
 We can now use `sort` to sort the lines in this file:
 
 ```bash
-$ sort lengths.txt
+$ sort lengths.txt -n
```
sida-wang, line 93: Can't replicate the results in the next code block without the option. Following lines of prose may need to be updated.

chapters/teams.Rmd
```diff
@@ -864,7 +864,7 @@ In a large project,
 a \gref{product manager}{product_manager} decides how important items are,\index{product manager}\index{project!product manager}
 while a \gref{project manager}{project_manager} is responsible for estimating effort\index{project manager}\index{project!project manager}
 and tracking progress.
-In a typical research software projects,
+In a typical research software project,
```
sida-wang, line 867: Can also be 'In typical research software projects,', happy to change based on feedback.

chapters/bash-tools.Rmd
DamienIrving: If you update the sentence following this command so that it says "we use `sort`'s `-n` option" instead of "we should use `sort`'s `-n` option" then that should be the only update the prose needs.

chapters/teams.Rmd
DamienIrving: I think this looks good the way you've done it.

DamienIrving (approved):
Thanks for picking up these typos, @sida-wang!

(Note to others: Sida and I are part of the same research team at UNSW and he has been reading through the book.)

If you go ahead and make the minor edit to the prose that I suggested in comments, I'll then go ahead and merge this PR :smile:

chapters/bash-tools.Rmd
```diff
@@ -106,7 +106,7 @@ $ sort lengths.txt -n
 ```
 
 Just to be safe,
-we should use `sort`'s `-n` option to specify that we want to sort numerically.
+we use `sort`'s `-n` option to specify that we want to sort numerically.
```
sida-wang, line 109: Updated as previously discussed.

PR #595, opened 2021-01-11
Fix error in titles.txt exercises

### LINKED ISSUE #594: Issue with exercise 3.8.7
Given `titles.txt` as 
```
Sense and Sensibility,1811
Frankenstein,1818
Jane Eyre,1847
Wuthering Heights,1847
Moby Dick,1851
The Adventures of Sherlock Holmes,1892
The Time Machine,1895
Dracula,1897
The Invisible Man,1897
```
from exercise 3.8.5, options 3 and 4 (listed below) from exercise 3.8.7 give the same answer. Here the goal is to generate a frequency table of publishing year. 
3. `cut -d, -f 2 titles.txt | uniq -c`
4. `cut -d, -f 2 titles.txt | sort | uniq -c`

However the solutions state that 3 is incorrect. The main issue here is that `titles.txt` is already sorted by year which means the sort in option 4 has no effect. Instead of updating the solution, it seems more appropriate to sort `titles.txt` alphabetically which would make option 3 incorrect.

DamienIrving:
Resolves #594

### DIFF
--- a/chapters/bash-tools.Rmd
+++ b/chapters/bash-tools.Rmd
@@ -899,18 +899,18 @@ you combine with it in a pipe to remove all duplicated lines?
 
 ### Pipe reading comprehension {#bash-tools-ex-reading-pipes}
 
-A file called `titles.txt` contains the following data:
+A file called `titles.txt` contains a list of book titles and publication years:
 
 ```text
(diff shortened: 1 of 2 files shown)


gvwilson:
LTGM

PR #599, opened 2021-01-13
Fix wheels content

### LINKED ISSUE #597: Wheel distribution explanation
Section 14.5 mentions two types of distributions: source distributions and wheel distributions. While source distributions are explained and have a glossary term, wheel distributions are mentioned once in the sentence
> We have now uploaded both types of distribution, allowing people to use the wheel distribution if their system supports it or the source distribution if it does not.

I think it might be worth having an extra sentence that explains what it is as well as a glossary term. My google searches suggest that its a built version of the source code which makes for a faster installation but I'm still a bit unsure.

DamienIrving:
Resolves #597

The text in the packaging chapter erroneously stated that we had uploaded both the source distribution and wheels for `pyzipf` to PyPI, when in fact we had only done the source distribution. This PR fixes this by adopting Option 1 described in #597.

### DIFF
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -810,7 +810,7 @@ which we can install with:
 ```
 
 Following the [Python Packaging User Guide][pypi-user-guide],
-we upload our distributions from the `dist/` folder
+we upload our distribution from the `dist/` folder
 using the `--repository` option to specify the TestPyPI repository:
 
 ```bash
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -843,13 +843,11 @@ In the exercises,
 we will explore additional metadata that can be added to `setup.py`
 so that it appears on the project webpage.
 
-We have now uploaded both types of distribution,
-allowing people to use the wheel distribution if their system supports it
-or the source distribution if it does not.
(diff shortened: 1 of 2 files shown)


sida-wang:
Minor point but should line 813 in `packaging.Rmd` be changed to 'distribution' from 'distributions'? Can't make the comment in the files changed code.

chapters/packaging.Rmd
```diff
@@ -897,6 +895,19 @@ and doesn't download them again.
 Once we are happy with our package at TestPyPI,
 we can go through the same process to put it on the main [PyPI][pypi] repository.
 
+> **Python Wheels**
+>
+> When we installed our package from TestPyPI,
+> the output said that it collected our source distribution 
+> and then used it to "build a wheel" for `pyzipf`.
```
joelostblom, line 902:
```suggestion
> and then used it to "build a wheel" for `pyzipf`.
> A wheel is a pre-built binary installation format,
> that is smaller and faster to install than the source file we uploaded,
> but not as versatile.
```

I'm suggesting a short high-level description of a wheel, something like this.

joelostblom (approved):
Minor suggestion, good with or without

PR #600, opened 2021-01-14
Address wheels comment from Joel

DamienIrving:
I didn't get a chance to address https://github.com/merely-useful/py-rse/pull/599#discussion_r557064922 from @joelostblom before #599 was merged. This PR basically takes Joel's suggested definition of a wheel and puts it in the glossary, since it felt a bit weird to talk about it being a "pre-built" thing when in the Python wheels call-out box we talk about the fact that pip built a wheel from the source distribution (i.e. in that instance it wasn't pre-built). Hope that makes sense.

### DIFF
--- a/chapters/packaging.Rmd
+++ b/chapters/packaging.Rmd
@@ -899,7 +899,7 @@ we can go through the same process to put it on the main [PyPI][pypi] repository
 >
 > When we installed our package from TestPyPI,
 > the output said that it collected our source distribution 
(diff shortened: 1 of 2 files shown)


joelostblom (approved):
LGTM, I hesitated to include "pre-built", I couldn't quite decide if it was helpful for novices or confusing because it is technically redundant when we also say "binary". We could just call it "built" or "binary" distribution, but this is minor so I am going to merge this.

PR #602, opened 2021-01-25
Biographies

gvwilson:
Please have a look at `taylor-francis/bios.txt` and add a few more words about yourself.


DamienIrving (approved):
Done :smile:

joelostblom (approved):

DamienIrving:
This is a minor thing and I don't mean to throw shade on R or the amazing R skills y'all have, but given that this is a Python book is there too much mention of R in these bios?

i.e. Should we replace some bits like "he develops R packages" with "he develops software packages"?

PR #616, opened 2021-03-14
Minor fixes for final proof

DamienIrving:
I've just looked over the final PDF document that @gvwilson sent around and spotted a few overflow lines and syntax highlighting issues.


lwjohnst86 (approved):
Likewise @DamienIrving :grin:

PR #628, opened 2021-08-04
Advertising blurb

gvwilson:
Adds `blurb.txt` with back-cover text for sharing on social media, and the PDF author flyer from CRC Press.


lwjohnst86 (approved):
Looks good!

PR #629, opened 2021-08-17
Add date and other info to Lin2020

patrickmineault:
The paper is no longer in press. This should update the citations from (Lin, Ali, and Wilson, n.d.) -> (Lin, Ali, and Wilson, 2020).


joelostblom:
Thanks!

PR #631, opened 2021-09-10
Fix typo in error message for directory copy

mikerenfro:
Tested with
```
% mkdir 1
% cp 1 2
cp: 1 is a directory (not copied).
```


joelostblom:
Thanks!

PR #636, opened 2021-09-13
Fix wrong md formatting in git exercise

uekerman:
Code blocks are apparently not allowed within a numbered list. Needs a new line to be formatted correctly.

Currently, it looks like [this](https://merely-useful.tech/py-rse/git-cmdline.html#git-cmdline-ex-history)

Btw, great book :+1: Will use it in our course next semester.


joelostblom (approved):
Thanks!

PR #637, opened 2021-09-13
Avoid implying that redirection is identical to a file parameter

mikerenfro:
Not the only way to adjust this wording, but the redirection operator is not always identical to including a command-line parameter for a filename. Works for `wc`, `sort`, and similar filtering commands, but not everything such as `ls`, `cp`, etc.


joelostblom (approved):
Thanks again!

PR #639, opened 2021-10-14
Adding basic web site statistics using plausible.io.

gvwilson:
https://plausible.io/ offers an ethical approach to collecting web traffic statistics. This addition will let us count the number of people visiting https://merely-useful.tech/ (we can then drill down to the py-rse book or others). I'm assuming that once this is merged, all of the HTML pages will be updated...

### DIFF
--- a/includes/header.html
+++ b/includes/header.html
@@ -1,1 +1,2 @@
 <link rel="shortcut icon" href="favicon.ico" />
+<script defer data-domain="merely-useful.tech" src="https://plausible.io/js/plausible.js"></script>


joelostblom (approved):
Yay for a considerate alternative to Google analytics!

k8hertweck:
I just added Plausible tracking to another project built in Jekyll, and it required a few lines be added to the config file.

I found [this bookdown project](https://github.com/oscarbaruffa/BigBookofR) that uses Plausible and seems to only include the code snippet in the header (as you've done), so fingers crossed we're good to go.

gvwilson:
@k8hertweck can you please point us at the config or summarize here? I _think_ adding to the include file will be enough, but as a backup...

k8hertweck:
@gvwilson That's the thing! I don't see any other place the config is necessary for a site rendered via bookdown. There isn't one in the link I shared (only the code snippet in the header).

lwjohnst86:
Let's try it out then! Merging and @gvwilson you can watch to see how it goes, if it works.

